PR #15, opened 2018-02-22
Auth flow update

rijkvanzanten:
This update styles the login screen, adds error / warning notices, and makes the auth flow more stable


rijkvanzanten:
screen shot 2018-02-22 at 5 08 09 pm

PR #21, opened 2018-02-26
Login style tweaks

benhaynes:


src/components/form/Input.vue
```diff
@@ -84,6 +84,8 @@ input {
 
   &:-webkit-autofill {
     box-shadow: inset 0 0 0 1000px var(--white) !important;
+    color: var(--dark-gray) !important;
+    -webkit-text-fill-color: var(--dark-gray) !important;
```
rijkvanzanten: `!important` ughh 🤢 

(Too bad it's needed to override the user agent styles)

PR #26, opened 2018-03-09
Style tweaks

### LINKED ISSUE #23: All the small things...
- [x] Global selection color, like 5% of primary color
- [x] Breadcrumb is 13px from left... could this be 20px?
- [x] Optical vertical alignment of breadcrumb. It seems like 1-2px too high because of descenders
- [x] Smaller Directus logo. Match 6.x
- [x] Loading blocker (smoke) should be 100% opacity and then fade in with global elements (header, etc). If you're fancy we can add an animation to the blocker :)
- [x] Breadcrumb separator (and active crumb) should have "normal" cursor (not text select)
- [x] Sidebar borders should be lightEST gray

_Update: remove sidebar nav_

rijkvanzanten:
🎉 

Closes #23


benhaynes:
* [ ] Responsive hamburger is double padded
* [ ] Global border radius should be 3px, not 4px
* [ ] Action button icons should be larger (all icons should use the recommended MD intervals, see below)


> Although the icons in the font can be scaled to any size, in accordance with material design icons guidelines, we recommend them to be shown in either 18, 24, 36 or 48px. The default being 24px.
http://google.github.io/material-design-icons/

Issue #52, opened 2018-03-26
Add support for error 2002 on login (no db connection)


rijkvanzanten:
When the API can't connect to the DB, it returns error `2002` to the user. As of right now, the login view assumes that every error returned by the API is a user error, but this doesn't have to be.

Based on the new api-docs, user errors can only be in the range x >= 100 < 200:

> Authentication error codes
0100 - Invalid Credentials (404)
0101 - Invalid Token (401)
0102 - Expired Token (401)
0103 - Inactive User (401)
0104 - Invalid Reset Password Token (401)
0105 - Expired Reset Password Token (401)
0106 - User Not Found (404)
0107 - User with a given email Not Found (404)

https://github.com/directus/api-docs/blob/2.x/errors/auth.md

rijkvanzanten:
I added a couple of API error codes to the translations already (`/lang/locales/en-US.js`), but not yet code 2002

I (wrongly) assumed that all error codes being served from the API would be because of a user error, so I made it in a way that it renders the yellow warning for api-returned errors and red errors when the api couldn’t be reached (code = -1)

BUT code 2002 returned from the error should also render in RED

Yellow = user error (something has to be fixed in the credentials -> wrong username / etc)
Red = server error (something has to be on the api -> missing config, wrong db creds, etc)

rijkvanzanten:
Has been closed by #67

PR #67, opened 2018-03-27
Add support for 2002 error code

honzabilek4:
#52 Type of the error is determined by the error code value. If 100<= x < 200 then it's an user error and it will be displayed in yellow (warning). Other codes are going to be red (error).


rijkvanzanten:
Just in, 2002 changed to `11`:

https://github.com/directus/api-docs/issues/19#event-1544346025

honzabilek4:
Changed the code number to 11.

PR #82, opened 2018-04-03
Live Latency Indicator

rijkvanzanten:
latency

Shows the signal strength in form of latency to the api. Polls every 5 seconds to `/ping` for updated latency stats. Keeps the last 200 pings in memory to be able to render a graph in settings (to be build)


theharshin:
Hey @rijkvanzanten Is there any config to disable this in a development environment without touching an actual code? I couldn't find any so can we add an environment check? The browser console and network tab are filled with it 😕

rijkvanzanten:
Good point. I wouldn't mind adding a `process.env.NODE_ENV` check to only enable this in production mode.

PR #268, opened 2018-04-24
Add activity page

### LINKED ISSUE #231: Add activity page
Would be an alias of `directus_activity`, but only showing _your_ items. This can be a normal tabular listing view with the following column setup:

* Action (custom interface, see below)
* Datetime (relative mode)
* Message

screen shot 2018-04-20 at 6 59 33 pm

rijkvanzanten:
Also adds "official" support for readonly listing view routes.

Closes #231


rijkvanzanten:
Note: the permissions / what's being viewed by users is based on the permissions to be implemented in a future PR.

Issue #251, opened 2018-04-23
Add auto-logout based on directus_settings value


rijkvanzanten:
It should log the user out after _n_ minutes as set in the DB

rijkvanzanten:
Closed by #347

PR #347, opened 2018-05-02
Log out after n minutes of inactivity

rijkvanzanten:
Automagically logs you out after _n_ minutes (as setup in `directus_settings`) when you don't actively use the platform.


rijkvanzanten:
Closes #251

PR #405, opened 2018-05-17
Create global event bus

honzabilek4:
Just a simple event bus as a Vue plugin.


benhaynes:
Thank you so much @honzabilek4 !!! This is awesome... and we really appreciate the help! 👍

PR #418, opened 2018-05-20
Add new notification plugin

honzabilek4:
I've removed the `vue-notification` plugin and added `Notyf` instead. Since it isn't a Vue component, we can now use it in our global event handler. 
It's going to need some improvements as suggested [here](https://github.com/caroso1222/notyf/issues/22), but it works for now. 
I've also added some basic event handler with a method for handling the [error event](https://github.com/directus/app/wiki/Error-handling).
@rijkvanzanten I'll be happy to hear your feedback.


benhaynes:
Thank you so much Jan! Excited to try out the new notifications system. 😄

src/main.js
```diff
@@ -66,6 +64,14 @@ Vue.use(VueTimeago, {
     nl
   }
 });
+Vue.component("draggable", VueDraggable);
+
+EventBus.on("error", error => {
+  EventHandler.handleError(error);
```
rijkvanzanten: Instead of connecting the EventHandler methods to the events in the main file, would it be possible to import the EventBus _in_ the EventHandler file? That way we don't have to manage the events in two places 🙂

src/assets/notyf.scss
```diff
@@ -0,0 +1,196 @@
+$red-notyf : #ed3d3d;
+$green-notyf : #3dc763;
+
+@keyframes fadeinup{
+  0%{
+    opacity: 0;
+    bottom: -15px;
```
rijkvanzanten: This needs to animate on the `transform` property instead of the `bottom` property, if possible.

src/assets/notyf.scss
```diff
@@ -0,0 +1,196 @@
...
+}
+
+.notyf__icon--alert{
+    &:after, &:before{
+      content: "";
+      background: $red-notyf;
+      display: block;
+      position: absolute;
+      width: 3px;
+      border-radius: 3px;
```
rijkvanzanten: Use `var(--border-radius)` instead

src/main.js
```diff
@@ -66,6 +64,14 @@ Vue.use(VueTimeago, {
...
+});
+EventBus.on("message", message => {
+  EventHandler.handleMessage(message);
+});
```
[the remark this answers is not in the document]
rijkvanzanten: I don't know if `message` is descriptive enough.. 

Lets remove it for now, and focus just on the error notification 👍

src/events/index.js
```diff
@@ -24,6 +25,10 @@ Object.defineProperties(EventBus, {
   }
 });
 
+EventBus.on("error", error => {
```
rijkvanzanten: This can be refactored to `EventBus.on("error", handleError)`

Not important, more a quick tip 😉

rijkvanzanten:
Lets not forget to update the styling. https://github.com/directus/app/issues/425

src/events/index.js
honzabilek4: Ah, thanks! :)

PR #564, opened 2018-07-27
Activity updates

benhaynes:
All good except my comment styles aren't working (?)


src/components/activity-overview/activity-overview.vue
```diff
@@ -21,22 +21,26 @@
       :key="activity.id">
       <i
         v-if="activity.type === 'comment'"
+        v-tooltip="'Comment'"
```
rijkvanzanten: No hardcoded pieces of text por favor. Instead, add a key to `en-US.json` and use `$t('comment')`

PR #619, opened 2018-08-08
Not working :(

benhaynes:


benhaynes:
The styling on the label for the layout chooser.

PR #640, opened 2018-08-16
Add German translation file

Zielgestalt:


benhaynes:
Thank you @Zielgestalt – we'll get this reviewed and merged asap!!

rijkvanzanten:
Awesome, thanks! 😄 

And here I was thinking Dutch was going to be the first officially supported second language in 7.. You beat me to it 😄 🇩🇪

rijkvanzanten:
@benhaynes low-key hoping nothing overflows and breaks with all the long German words 🙃

Zielgestalt:
I tried to avoid overflows. But we have some in the settings overview page (in the cards). But I think it's still understandable.
image
Perhaps you can give the cards a bit more width. I think it's ok for German, so it isn't urgent. But French is even longer than German. Perhaps you can keep that in mind.

benhaynes:
The card component is based on one of our standard widths, the next size up being more than double... which would be too wide. I'll keep trying to think of a solution, but we do have a tooltip for any overflow text in the meantime.

PR #683, opened 2018-08-23
Override the sort ghost style

benhaynes:


rijkvanzanten:
If you click the drag handler, the whole row disappears. It only reappears when actually dragging

benhaynes:
OK, I updated the style... annoying not being able to use my original solution, but this should be fine.

PR #820, opened 2018-09-12
Details border style

benhaynes:


src/components/details.vue
```diff
@@ -91,12 +91,13 @@ export default {
     }
   }
 
-  .content {
-    padding-bottom: 30px;
-  }
+  // .content {
```
rijkvanzanten: Lets not have commented out code blocks in the source code. We have Git version control to go back and forth between revisions.

PR #839, opened 2018-09-20
fixes #837

### LINKED ISSUE #837: User Directory Breadcrumb
Clicking the User Directory breadcrumb takes you to: `/collections/users` ... a 404

It should go to `/users`

benhaynes:


src/routes/item.vue
```diff
@@ -231,17 +231,27 @@ export default {
...
-      ) {
+      if (this.collection === "directus_users") {
         return [
           {
             name: this.$t("user_directory"),
-            path: "/collections/users"
+            path: "/users"
+          },
+          {
+            name: (this.primaryKey == this.$store.state.currentUser.id) ? this.$t("editing_my_profile") : this.newItem ? this.$t("creating_item") : this.$t("editing_item"),
```
rijkvanzanten: This is not very legible.. Lets try to clean this up before we merge it in.

rijkvanzanten: A single ternary can improve legibility, a nested one is always harder to read:

```js
// Good
const result = condition ? a : b;

// Bad
const result = condition ? a : b ? c : d;
```

benhaynes: I know, but I only have so much knowledge/time... just trying to get things working/done 😕

Issue #881, opened 2018-09-29
Format Role Name


benhaynes:
Settings>Roles...
Admin and Public are stored with case... but should we run these role names through the title formatter?

benhaynes:
![screen shot 2018-09-29 at 12 19 09 pm](https://user-images.githubusercontent.com/522079/46248031-e44b5400-c3e1-11e8-8bd7-0c5626c3c91a.png)

PR #924, opened 2018-10-05
#881: Format Role Name

gitofwebmobiledev:


rijkvanzanten:
Ah, I'm assuming it's a fix for https://github.com/directus/app/issues/881

Do we want to force create the group name with an uppercase and display the raw DB value or store whatever and title-format the raw value on display?

Thoughts @gitofwebmobiledev @benhaynes ?

gitofwebmobiledev:
@rijkvanzanten yes, I fixed #881.
And, we need to create the group name with capitalize

rijkvanzanten:
> And, we need to create the group name with capitalizing

Why?

PR #937, opened 2018-10-08
fixed bug with sorting fields

isntlazy:
Please check. Now the fields is in right order even if you just created new collection with fields and did a page refresh for example


rijkvanzanten:
Hey @isntlazy! Thanks for the PR.

`a` and `b` are the full field info objects, so sorting by `a - b` doesn't actually do anything in this case. It works, because the API started returning the fields based on the `sort` order in https://github.com/directus/api/commit/98b23211879197b9911cc09e726c454821cba065.

Issue #936, opened 2018-10-08
Login: Don't show API as dropdown if only one option


benhaynes:
If the App config only has one API option then the login page shouldn't show it as a dropdown... it should just be a static string.

![screen shot 2018-10-08 at 1 38 47 pm](https://user-images.githubusercontent.com/522079/46624682-8ca69a00-caff-11e8-9b2a-0f8ca44fb753.png)

PR #945, opened 2018-10-08
Only display dropdown when more than one option is available

N1ck:
Resolves: https://github.com/directus/app/issues/936

If the App config only has option, it won't display it as a dropdown:
image

Let me know if there's anything I missed 🙂

### DIFF
--- a/src/routes/login.vue
+++ b/src/routes/login.vue
@@ -11,7 +11,7 @@
         <h1 v-else>{{ resetMode ? $t('reset_password') : $t('sign_in') }}</h1>
 
         <label class="project-switcher">
-          <select v-model="selectedUrl" :disabled="loading">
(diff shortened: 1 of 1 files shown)


benhaynes:
Thank you @N1ck !! This looks good to me, but I'll wait to hear from @rijkvanzanten.

We really appreciate the PR!! ❤️

benhaynes:
@N1ck @rijkvanzanten ... just a question (I haven't tested this yet). Will this work if you have `other` APIs enabled? That wouldn't be in the array but it needed to show the input for typing in an API URL.

rijkvanzanten:
Ahh yes @benhaynes good point. I'll update so it will still render the dropdown when you have the allowOther option checked

Issue #939, opened 2018-10-08
Interface: Divider>Options – Divider Text


benhaynes:
When you try to delete the default text in `divider-text` or `divider-description` it resets the default text when the value is empty. If you delete the text, it should stay empty. 

Also, it's trimming whitespace off the end of the value. (If you have `testing 1` and delete `1`... it also deletes the space).

Both of these issues make it very easy to make a mistake when entering values.

![screen shot 2018-10-08 at 2 59 27 pm](https://user-images.githubusercontent.com/522079/46628232-e496ce00-cb0a-11e8-9227-0002345c3dd7.png)

rijkvanzanten:
The reason that the interface is reset with the original value is because it falls back to it:

https://github.com/directus/app/blob/5d5656d338fe2603daf4f23fe710a7d9f2c9e439/src/components/field-setup.vue#L239

The backspace-issue is caused by the `trim` option:

https://github.com/directus/api/blob/3bd595cb308c024ddad18c8d7c7a2974f422e55b/extensions/core/interfaces/text-input/input.vue#L25-L27

Trimming works just fine when writing, but when back-spacing, it feels very unnatural. This trimming should only happen when the value's length increases, not decreases, so it only trims on adding characters.

benhaynes:
In my opinion the trim should only happen on Save, if that's possible

rijkvanzanten:
It's not @benhaynes, interfaces can send up values to the parent, but are completely agnostic as to what happens to that value.

benhaynes:
Gotcha. Then checking if you're typing forward or "backwards" seems like the next best solution.

PR #951, opened 2018-10-09
add placeholder text

segunolalive:
Fixes issue [939](https://github.com/directus/app/issues/939)

Allows users to clear default text in input fields, then displays the default text using input placeholder if  left empty

### DIFF
--- a/src/components/extension/interface/input/input.vue
+++ b/src/components/extension/interface/input/input.vue
@@ -83,6 +83,10 @@ export default {
     values: {
       type: Object,
       default: null
+    },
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
Hey @segunolalive!

Thanks for the PR, unfortunately, this breaks more than it solves.. This fix changes every single interface. Not every interface supports placeholders and not every field has a value that's suitable for a placeholder. 

The symptoms in #939 are caused by two smaller issues that can be fixed individually without affecting the whole application. See the issue (#939) for more info on that.

Issue #929, opened 2018-10-05
Error going from detail to Single collection page


benhaynes:
1. Go to an Item Detail page
1. Click on a SINGLE collection in the nav...
1. Get an error... `Server Trouble`

segunolalive:
This error is due to the wrong query parameters being sent along with the request.

This is the error
`class: "Directus\Database\Exception\InvalidFieldException"
code: 202
file: "/var/www/html/src/core/Directus/Database/TableGateway/RelationalTableGateway.php"
line: 2012
message: "Invalid field "name""`

It appears that the fields for the previous details page are being appended to the next details page. I haven't been able to tell where that is being done.

@rijkvanzanten any pointers here?

segunolalive:
@benhaynes @rijkvanzanten take a look at the fix in the PR above. I'd love your feedback.

rijkvanzanten:
Hey @segunolalive! Your assessment was right, but unfortunately, we can't strip out the whole query params. It's using those to fetch just the fields we need and fetch the nested relational items as well. Stripping all of that would cause the edit form to fail for all relational fields. 

This way of fetching the query params has worked for navigating in "fresh" (eg browse -> detail) but not on "update" (eg detail -> another detail). In Vue, this is the difference between `beforeRouteEnter` and `beforeRouteUpdate` which makes me think that the state of the original detail page doesn't fully get reset.

segunolalive:
yes, the state isn't being cleared perhaps because it's reusing the `item.vue` component for performance

rijkvanzanten:
> so, would it be appropriate to move that logic into beforeRouteEnter and make the network request within the callback passed to next?

I'd like to keep the request inside the [route navigation guard](https://router.vuejs.org/guide/advanced/navigation-guards.html), that way the loading flow feels more "natural" to the user (eg, it's loading until done and then rendering).

The problem here is that `this.collection` in `beforeRouteUpdate` is using the old collection name. `this.collection` is passed in by `vue-router` and is the current collection in the URL. However, as the name suggests, _before_ RouteUpdate fires before the route is updated. Therefore, `this.collection` is still the "old" collection. 

We're already fetching the correct "new" collection from the `to` param in `beforeRouteUpdate`:

```js
beforeRouteUpdate(to, from, next) {
    const { collection, primaryKey } = to.params;
```

But we're not using that `collection` variable where we should:

```js
    return this.$api
      .getItem(collection, primaryKey, {
[...]

PR #953, opened 2018-10-09
fetch fields for the correct collections

segunolalive:
fixes isssue #929


rijkvanzanten:
Hey @segunolalive! Your assessment was right, but unfortunately, we can't strip out the whole query params. It's using those to fetch just the fields we need and fetch the nested relational items as well. Stripping all of that would cause the edit form to fail for all relational fields. 

This way of fetching the query params has worked for navigating in "fresh" (eg browse -> detail) but not on "update" (eg detail -> another detail). In Vue, this is the difference between `beforeRouteEnter` and `beforeRouteUpdate` which makes me think that the state of the original detail page doesn't fully get reset.

segunolalive:
yes, the component state isn't being cleared perhaps because it's reusing the `item.vue` component for performance

rijkvanzanten:
Ref https://github.com/directus/app/issues/929

segunolalive:
PR updated @rijkvanzanten

Issue #947, opened 2018-10-08
Creating Item Page Title


benhaynes:
The page title (as seen in browser tab) is saying "Editing XXXX" when creating an item. Page titles should say:

* Creating: `Creating Item: [collection-name]`
* Updating: `Editing Item: [collection-name]`
* Batch Updating: `Batch Editing Items: [collection-name]`

tthompson899:
I'd like to work on this issue

benhaynes:
Hey @tthompson899 – that sounds great, thank you! Consider it yours!

rijkvanzanten:
@tthompson899 This is the bit of code that sets that title:

https://github.com/directus/app/blob/d14102ee12010e798ecee5d321e8feec23c322a8/src/routes/item.vue#L183-L193

As you can see, it's always returning the translation for "editing".
`this.$t` is using this translation: https://github.com/directus/app/blob/d14102ee12010e798ecee5d321e8feec23c322a8/src/lang/locales/en-US.js#L186

The fix for this would be to add a few extra conditionals in that `metaInfo` method above that returns other strings if the item is new or batch. You should be able to use `this.isNew` and `this.batch` to check for that 😄

tthompson899:
@rijkvanzanten Good to know! I was looking in that file so I was close 😄

tthompson899:
Ready for review.

PR #960, opened 2018-10-12
fixed page titles with correct verbage for create, edit and batch

tthompson899:
fixed [issue 947](https://github.com/directus/app/issues/947)


rijkvanzanten:
Heya! You accidentally removed the example config

benhaynes:
Thanks @tthompson899 !!

The `create` key in translations might be used in other places where it needs to be super short to fit in its container. Did you do a quick check to make sure that's the only place that uses it? Instead of changing that translation you should create a new one so we don't break things in other places.

tthompson899:
All of these have been addressed in the latest commit - Let me know if there's anything else that needs to be fixed.

rijkvanzanten:
Thanks @tthompson899! 😄

PR #1005, opened 2018-10-19
Fixes #973 Tile Layout: Exit selection mode after batch delete 

### LINKED ISSUE #973: Tile Layout: Exit selection mode after batch delete
After you delete an item(s) from the Tile layout (eg: File Library) you are still in selection mode (with the checkbox circles in the corner). We should exit this mode and return to the normal view.

ShadowRonin:
Fixes #973. Selection mode is not exited after delete.


rijkvanzanten:
Thanks @ShadowRonin! 😄

PR #1050, opened 2018-10-28
Datatype displayed empty value & caused issues

shealavington:
Fixed issue: Some fields (Relational) were throwing an error about having a length and not needing one, after some debugging I found that the `datatype` was being set to the same value as the `type`, therefore submitting datatype as empty on update. I fixed that and the value came through, however due to the datatype needing to be compared as uppercase, it's now compared as uppercase. (Closing issue "Updating a "Many to one" field #530" in API)


benhaynes:
Thanks @shealavington !!

PR #1051, opened 2018-10-28
Settings>Roles - field order

shealavington:
This reorders the permissions page to have name and description show above permissions as per the request; https://github.com/directus/api/issues/444


benhaynes:
Thanks @shealavington! I'll wait for Rijk to review... but this looks good to me!

src/lang/locales/nl-NL.js
```diff
@@ -140,5 +140,7 @@ export default {
   version: "Versie",
   version_and_updates: "Versie en updates",
   view_type: "Bekijk type",
-  yes: "Ja"
+  yes: "Ja",
+  duplicating_field: "Duplicaat veld",
```
rijkvanzanten: Ik denk dat het grammaticaal iets netter is als we zeggen "Dupliceer veld" 😉

shealavington: 😆 Google didn't suggest that, haha!

rijkvanzanten:
> This reorders the permissions page to have name and description show above permissions as per the request

The original request by @benhaynes was:

> Name + Description should be directly below the permissions interface... right now it is below the IP Whitelist.

So I'm afraid this is not fixing the original issue 😬

PR #1052, opened 2018-10-28
WIP: Notifications

honzabilek4:
Hi,
I've tried to implement better notification system as we discussed. It includes a new module in store where all the notifications are kept and two components: `notifications.vue` and `item.vue` which handle the displaying of items. There's also a `notifications.js` file, which allows me to mutate the store and create a notification from anywhere in the app.

**Usage example:**
```js
this.$notify({
        title: "New mention",
        details: "Click to see the comment",
        iconRight: "arrow_forward",
        iconMain: "message",
        type: "info", // info / success / warning / error or undefined (general)
        delay: 0, // defaults to 5000ms; 0 is permanent
        clickCallback: () => {
          this.$router.push("/activity");
        }
      });
```

**Preview:**

screenshot 2018-10-28 at 01 31 39

**TODO:**
* Add markdown support.  (I'm going to add it asap using the `vue-markdown` library, should be a quick change)
* Support for progress bar (I wanted to use the `progress-ring.vue` we already have, but it's too tightly coupled and I didn't wanted to repeat the logic, so we might need to discuss it)

**Areas for improvement:**
* Review the naming
* Making the leave animations smoother
* Usage simplicity?

This pull request is WIP, It'll need some modifications before actually merging it. Also the `$notify` calls throughout the entire app will have to be changed.


rijkvanzanten:
> Add markdown support. (I'm going to add it asap using the vue-markdown library, should be a quick change)

`snarkdown` is included as a helper. You can convert any markdown to html in any part of the system using `this.$helpers.snarkdown("# hello world");` 😄 

> (I wanted to use the progress-ring.vue we already have, but it's too tightly coupled and I didn't wanted to repeat the logic, so we might need to discuss it

What do you mean by too tightly coupled? Can we uncouple it enough for you to use?

rijkvanzanten:
The code is already looking really good @honzabilek4! 😄 Amazing work!

honzabilek4:
> `snarkdown` is included as a helper. You can convert any markdown to html in any part of the system using `this.$helpers.snarkdown("# hello world");` 😄

Good news, I didn't notice it before 🙂 Then it's even easier to do. 
Do you want also the title to be in markdown? Or only the additional information? Both options are okay to me.
 
> What do you mean by too tightly coupled? Can we uncouple it enough for you to use?

Maybe I just need to think a little bit more about how the progress component should work together with the notifications. There are multiple solutions to this, but it might be easier to show you the problem on screen, than trying to describe it here 🤓

benhaynes:
I think only the body text should support markdown... not the title. I defer to @rijkvanzanten on the progress component.

rijkvanzanten:
> I think only the body text should support markdown... not the title

Agreed!

honzabilek4:
I've added the support for markdown, the html elements will have to be styled as needed.

src/components/notifications/item.vue
```diff
@@ -0,0 +1,138 @@
...
+  computed: {
+    detailHtml() {
+      return this.$helpers.snarkdown(this.item.details);
+    }
+  },
+  methods: {
+    ...mapMutations("notifications", [REMOVE_NOTIFICATION]),
+    startItemTimeout() {
+      if (this.item.delay !== undefined && this.item.delay > 0) {
+        setTimeout(this.removeItemFromStore.bind(this), this.item.delay);
```
rijkvanzanten: Let's change this to an inline arrow function. 

```suggestion
        setTimeout(() => this.removeItemFromStore, this.item.delay);
```

src/components/notifications/item.vue
```diff
@@ -0,0 +1,138 @@
...
+      }
+    },
+    removeItemFromStore() {
+      this.$store.commit(REMOVE_NOTIFICATION, this.item.id);
+    },
+    actionClick() {
+      if (!(this.item.clickCallback instanceof Function)) {
```
rijkvanzanten: Let's rename `clickCallback` to `onClick`.

src/components/notifications/item.vue
```diff
@@ -0,0 +1,138 @@
...
+
+.title {
+  font-weight: 700;
+}
+
+.details {
+  color: var(--light-gray);
+}
+
+.error {
```
rijkvanzanten: Would it be possible to make `color` an option on `item` and have the code automatically use 100 and 500?

That way we can render the notification in any color 😄

src/components/notifications/notifications.vue
```diff
@@ -0,0 +1,47 @@
...
+  </div> 
+</template>
+<script>
+import VItem from "./item.vue";
+export default {
+  name: "VNotifications",
+  components: {
+    VItem
+  },
+  data() {
```
rijkvanzanten: Remove unused data method

src/notifications.js
```diff
@@ -0,0 +1,17 @@
...
+import { PUSH_NOTIFICATION } from "./store/mutation-types";
+
+let counter = 0;
+
+const defaultOptions = {
+  delay: 5000
+};
+
+const notify = function(params) {
+  let options = Object.create(defaultOptions);
```
rijkvanzanten: We could replace this with

```js
const options = {
  ...defaultOptions,
  ...params,
  id: this.$helpers.shortid.generate()
};
```

src/notifications.js
```diff
@@ -0,0 +1,17 @@
...
+  options = Object.assign(options, params);
+  options.id = counter++;
```
rijkvanzanten: We could alternatively use [shortid](https://docs.directus.io/app/contributor/helpers.html#shortid) which is available from `this.$helpers`

honzabilek4: This is definitely a better idea 🙂

src/components/notifications/item.vue
honzabilek4: Yess, thanks! 🙂 I did it the old way, haha 😆

src/notifications.js
honzabilek4: It's not possible to access `this` in plain javascript file, so I'll import `shortid` instead.

honzabilek4:
I've replaced all of the old notify calls with the new syntax. Can you please do a review and briefly test it, so we can merge this into master? 🙂

rijkvanzanten:
Thanks @honzabilek4, amazing PR ❤️

PR #1056, opened 2018-10-30
Duplicatable fields

shealavington:
There's now a three dotted menu on the right side of the collection fields under settings, there's an option here called Duplicate which allows you to now duplicate the entire field into another collection or the same collection, it saves everything except the field-name which has to be chosen. There's a blacklist of types that can't be duplicated. (Primary/Sort/Any Relational)


rijkvanzanten:
Thanks @shealavington! 😄

shealavington:
> Thanks @shealavington! 😄

My pleasure, I also saw last night there's a VueUI "Fix" script that I can run next time before committing, to fix code to the standards, cheers for fixing it up haha, I'll also come to you next time I need translation help haha 😛

PR #1101, opened 2018-11-07
Add compression after build

Anthonyzou:
Hi,

This adds a gulp script to precompress files after the build process, with gzip and brotli.


rijkvanzanten:
Hey @Anthonyzou! Thank you very much for the PR. Is there any way we can introduce the same functionality in the existing build chain by using the `vue.config.js` `configureWebpack` key? I'd like to keep the build chain all in the same system, otherwise it's going to be quite hard to keep track of what goes where.

Anthonyzou:
I updated it to use webpack plugins.

rijkvanzanten:
As far as I can tell, this won't work out of the box on any server, so I'll close it for now.

PR #1129, opened 2018-11-18
Add placeholder for disabled length field

### LINKED ISSUE #1070: Primary Key Datatype and Length
The default primary key (`id` as INT) is not looking good when you open it up:

screen shot 2018-11-05 at 3 05 15 pm

This should say `INT` as the MySQL DataType, and even if we're pulling the length from the type, we shouldn't have totally empty fields like `length`. It could show `15` (like MySQL clients), or some placeholder text that says it comes from the datatype.

shevaroller:
Adds placeholder to a disabled Length field in Settings -> Collection & Fields. Please see screenshot attached below.

1070-example

Closes #1070

### DIFF
--- a/src/components/field-setup.vue
+++ b/src/components/field-setup.vue
@@ -134,6 +134,7 @@
                 @input="length = $event;"
                 :value="lengthDisabled ? null : length"
                 :disabled="lengthDisabled"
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
Thanks @shevaroller!

Issue #1093, opened 2018-11-07
Non-blocking info sidebar


benhaynes:
Currently the Info Sidebar (the right one) opens a blocker over the page content when open. This makes it difficult to see (or interact with) content and activity/comments at the same time. It also marginalizes the Comments/Activity... so they are seen/used less. This is a very useful section of info, and as we improve its features we want to allow users to discover/use/see it more.

Our content forms max out at 620px, plus 200px for the nav sidebar, plus 350px for the info sidebar... is 1170px. At this size and above, the info sidebar could default to open without losing screen real-estate. I propose we remove the smoke and have the info sidebar auto-open (on Item Detail) when the viewport is over 1170px wide.

benhaynes:
Yeah, that's how it worked in version 6 and I'd like to do something similar. But I believe that Rijk had said there was some sizing issue with that in version 7.

Will look into it once all of these bugs are resolved

ceruulean:
Hey, I was able to implement the auto-display infobox and disable the blocker when the viewport is over 1235px (main content is 680px, sidebar is 360px, nav is 180px, padding is 20px, so really 1240px but I wanted some leeway)

I'm not sure what the content pushing would look like, since I have not used v6. Selecting 'Comments' already sorts comments from activity, so I don't see a need to do anything there? Is it just visual flavor when transitioning through the tabs?

rijkvanzanten:
@ceruulean I think @benhaynes means that instead of opening the info sidebar on top of the content, the content gets "pushed" in a little bit to make room for the sidebar. That being said, I'm totally in favor of showing it at all times if your screen real estate allows it. @benhaynes thoughts?

rijkvanzanten:
@benhaynes The main problem we had with the "push" was performance. Transitioning the push is near impossible to do, since there are like 40 reflows and draws happening in the browser in the transition time ~300ms. This will jank up even high-end computers.

benhaynes:
We can "pop" the sidebar in instead of transitioning it (maybe that helps?).

Yeah, all I really care about is showing the sidebar when the real-estate allows for it. Useful features in there that shouldn't be hidden if there's room to show them!

benhaynes:
@ceruulean — did you have a PR for this or just got it working locally?

Nitwel:
Another idea would be, to open it in a seperate tab or window of the browser, in this way, the sidebar is 100% seperated. But personaly, this seems a bit odd, doing it this way.

benhaynes:
[...]

PR #1158, opened 2018-11-29
Non-blocking info sidebar #1093

ceruulean:
Info sidebar auto-opens on item page and removes the blocker when wide enough
Also sets z-index of nav blocker because it was showing below the text area field


benhaynes:
Thanks @ceruulean!

src/components/sidebars/info-sidebar.vue
```diff
@@ -25,6 +25,12 @@ export default {
       default: false
     }
   },
+  mounted: function() {
```
rijkvanzanten: Lets use this syntax for methods to be more in line with the rest of the codebase:

```suggestion
  mounted() {
```

src/components/sidebars/info-sidebar.vue
```diff
@@ -25,6 +25,12 @@ export default {
...
+    var widthEnough = (window.innerWidth || document.documentElement.clientWidth || document.body.clientWidth) > 1235;
```
rijkvanzanten: Can we think of a way where we can keep the responsive design in CSS only? This will only run on mounted (once) so won't play nicely when the user is resizing the browser.

src/components/sidebars/info-sidebar.vue
```diff
@@ -25,6 +25,12 @@ export default {
...
+    if (this.$route.params.primaryKey != undefined && widthEnough) {
+      this.$store.commit(TOGGLE_INFO);
```
rijkvanzanten: I think we might have to refactor the info-sidebar to have this functionality in it by default.

@benhaynes already raised the question if it should be always available in the browse views as well. I'm almost starting to think that it doesn't really make sense to have a `info-sidebar` component separately to begin with? The functionality is starting to differ a lot between the browse and detail views.

What are your thoughts on removing the info-sidebar component and using a regular `<aside>` instead @ceruulean? That way we should be able to do most in CSS and clean up the logic quite a bit. (KISS)

ceruulean:
I didn't add a resize event because I didn't know what to put inside it, haha...
I assumed the sidebar should be the user's choice to make it pop up or not.
If the page loads without the sidebar, should the infobar automatically show after increasing the viewport? 
If the viewport is sized down, the blocker appears and the user can exit the sidebar. Sidebar is now "inactive," so does the App have to reactivate it? What if the user had closed it intentionally after the page loaded?

Maybe it's better to enable the Info button only when the viewport is too small, otherwise disable it.

ceruulean:
@rijkvanzanten Ok, this time no mounted function!
* Enable "Info" button only when the viewport < 1235px
* Info sidebar automatically shows when over 1235px on Item Detail page
* "Info" on Browse Item has been changed to "sort" but still functions as before
* blocker is enabled when viewport < 1235px

PR #1164, opened 2018-12-02
Improve table name and column name translation.

Chris533:
t1
t2
t3


rijkvanzanten:
@Chris533 Thanks for the new locale! About the system field names though, those aren't in the translation files themselves (which is also why they're not translated in runtime) but are instead in the database in the "translation" column.

However, seeing your fix made me think that it actually might make sense to have the system collections and fields translated in the main translation files, so we can make sure they're always in the right language no mattere what language the user is using.

Chris533:
@benhaynes @rijkvanzanten 

😁Thanks for your reply. I see, I will keep learning your future updates, your creation is really great amazing, and it's so user-friendly design. 👍👍👍

PR #1167, opened 2018-12-04
Infobar split aside

ceruulean:
Sorry, I screwed up my branches...

Submitting a PR for Non-blocking info sidebar #1093 but it's gotten more involved

If any decisions are questionable feel free to change or ask, I tried to do it with no extra javascript


rijkvanzanten:
This is looking really good, but there's one small issue. The `<transition>` element in `info-sidebar.vue` can only have 1 element in it. If look closely, you can see that the enter and leave animation isn't working when opening and closing the sidebar on smaller screens on the edit view

src/components/sidebars/info-sidebar.vue
```diff
@@ -1,11 +1,28 @@
...
+        v-if="active && !itemDetail"
+        :class="{ wide }"
+      >
         <div class="system"><slot name="system" /></div>
         <slot />
       </aside>
+
+      <aside
+        class="info-sidebar"
+        v-if="itemDetail"
```
rijkvanzanten: Because we're rendering this based on the media query now instead of using the v-if, the in and out transition isn't working for the sidebar on the item edit view

ceruulean: Ok, I added a css class to the aside so resizing transition works. I took this <aside> outside of <transition> since it never used vue's transition system in the first place. Is this the transition you were referring to?

rijkvanzanten: I think so! I'll confirm asap :)

PR #1168, opened 2018-12-04
Fix UserRoles

assmith:
Hi there,
               This PR fixes a bug in the role check logic for non-admin users when selecting checkboxes in tabular layout. Currently the test for showing the batch edit and delete buttons breaks trying to iterate on the user object instead of the roles array.


benhaynes:
Thanks @assmith! Good catch! :)

PR #1171, opened 2018-12-05
Portuguese translation

fernandoduro:
It\'s just a first version of the file, any sintaxe fixes that i find i will re-submit ;)


benhaynes:
Thank you @fernandoduro!! What an awesome contribution! 🇵🇹

PR #1197, opened 2018-12-13
Default fields on collection creation

### LINKED ISSUE #897: Create Collection: Default Fields
When creating a new collection through Directus, the dialog should have a few options to make system field setup easier. I see these as checkboxes for boilerplate fields:

`Automatically create these fields:`
* `ID` (default: on) Disable this since we need it on, it's just to show the user we're adding it
* `Status` (default: off)
* `Sort` (default: off) Should be set to hide on input
* `Created By` (default: on) Should be set to hide on input/browse, field should be readonly
* `Created On` (default: on) Should be set to hide on input/browse, field should be readonly

If these fields are added, they should also include a "Note" in `directus.fields` describing what they do.

c-ciobanu:
This pr add the option to select some default fields during the creation of a collection.

preview

Before i add the notes i want to know if i should localize them.
The labels too so everything will be localized.

Fixes #897.


benhaynes:
Thank you @c-ciobanu !!! This is awesome! 😄 

@rijkvanzanten has today off so we will review this first thing next week!

Can we think of another auto field? It'd be great if the layout were:

```
id             xxxx
status          sort
created_ by    created_on
modified_ by    modified_on
```

benhaynes:
If we can't think of another good initial field then maybe we should remove the hardcoded `id` so that this fits better. Thoughts @c-ciobanu ?

c-ciobanu:
I was thinking about removing the sort field since i don't thinks is used very often.
But maybe removing the id one is the best solution since it's required by default.
I think the id field should be created automatically when you call the /collections route without passing it from the app.

benhaynes:
Gotcha. Let's remove the `id`. In the future we can add a header for the primary key that let's you set the name and datatype... but first we need to make sure the API supports that, so let's wait.

For now, this is what we'd have:

```
status          sort
created_ by     created_on
modified_ by    modified_on
```

Let's have `status`, `created_ by`, and `created_on` **ON** by default (the rest off).

PR #1207, opened 2018-12-18
Warning on uploading a file larger then max file size

### LINKED ISSUE #1201: Warn users if exceeding max file size before upload
We know the max upload size (from the API) and should give a friendly warning _before_ the upload begins. The warning should also explain _why_ the upload won't work: `Your server is not configured to handle uploads of this size.`

c-ciobanu:
screenshot from 2018-12-18 22-05-24
A warning toast appears when you try to upload a file larger then max file size.

Fixes #1201.


src/components/upload.vue
```diff
@@ -113,6 +113,15 @@ export default {
       const formData = new FormData();
       const { name, size, type } = file;
 
+      if (size !== -1 && size > this.$store.state.serverInfo.maxUploadSize) {
+        this.$events.emit("warning", {
+          notify: this.$t("upload_exceeds_max_size", { filename: name }),
+          color: "orange"
```
rijkvanzanten:
```suggestion
          color: "warning"
```

Can we change this to use the warning color? That will make sure it's consistent if the user changes the warning color to something !== orange 🙂

c-ciobanu: Sure, I didn't know it was possible to customize the warning color.

Issue #636, opened 2018-08-15
Merge file-library.vue and items.vue


rijkvanzanten:
They're too similar to warrant the duplicate code and maintenance hell

c-ciobanu:
Hey, i would like to take this one.

rijkvanzanten:
Sounds good @c-ciobanu! I gotta be honest, it's a little bit 🍝 code. Sorry 'bout that!

c-ciobanu:
np, I was thinking about passing a prop to the /files route and then merge based on that.

benhaynes:
Thanks @c-ciobanu!!! This will make both pages more powerful and streamlined.

@rijkvanzanten is User Directory still using the core `items.vue` file?

rijkvanzanten:
As a matter of fact, with the feature requests for dropping on this page, combined with the other exceptions that files have (other title, custom new-item functionality), it might be better to keep them separated after all

benhaynes:
I'm down to discuss, but my thinking for "global drag and drop" was that _anywhere_ on the platform will accept files — but you determine the scope based on where you drag it. eg: drag on interface uploads and relationally adds to that field... drag onto random page and it just uploads to File Library.

PR #1211, opened 2018-12-21
file-library and items merge

c-ciobanu:
It's almost done, only one little problem that i can't seem to resolve.

When you login and land on the /files route all works fine but if you visit a collection then you come back the api call to retrieve the files fails and the files are not shown.

This is the error message:
screenshot from 2018-12-21 23-23-15
It seems like someone is passing an unwonted sort parameter to the call, I couldn't find where this happens.
screenshot from 2018-12-21 23-25-18
This only happens with certain collections (e.g. kiosks, members, movies).

@rijkvanzanten could you give me a hand with this?

Issue #636


benhaynes:
Thank you @c-ciobanu!!! @rijkvanzanten is on vacation this week but we'll review (and hopefully merge) this next week! ❤️

src/routes/items.vue
```diff
@@ -238,22 +275,27 @@ export default {
...
+        : this.$route.params.collection;
     },
     emptyCollection() {
-      return (this.meta && this.meta.total_count === 0) || false;
+      return this.meta && this.meta.total_count === 0;
     },
     filters() {
-      if (!this.preferences) return [];
-      return this.preferences.filters || [];
+      return this.preferences && this.preferences.filters
```
rijkvanzanten: I used to prefer smart one-liners over more "dumbed down" solutions, but IMO the old one is easier to read 🙂

c-ciobanu: Yeah, i can see your point, will change them back.

rijkvanzanten:
This is all looking good @c-ciobanu 😄 I haven't been able to figure out why it's not working when you navigate away from some of the other collections though 🤔

c-ciobanu:
@rijkvanzanten just found the problem, it was due to the fact that files is a different route and the `beforeRouteLeave` was not triggering.
Will push this and the other fixes soon.
We should change the name of this file in items-files or something else since i noticed if you try to debug the items file in components it will bug out the debugger and let you see the items file in routes.

rijkvanzanten:
I just pushed another fix for that issue @c-ciobanu 😄 

> We should change the name of this file in items-files or something else since i noticed if you try to debug the items file in components it will bug out the debugger and let you see the items file in routes.

Good point!

c-ciobanu:
@rijkvanzanten had to revert your fix because it was causing another problem on the users page.
Adding a `beforeRouteLeave` resolved it and now all seems to work fine, you should take a look too and see if it's all good.

rijkvanzanten:
There is a big refactor on the horizon that will take care of most of this change, and there have been a bunch of updates on both the item listing and the file listing separately, soo.. While I really really appreciate your efforts @c-ciobanu, I'm afraid I'm going to have to close this as we don't have the time to make it up to date nor be able to get it in an easily maintainable state. Also, I'm not too sure if all the extra conditionals are making things easier, or harder to maintain..

PR #1222, opened 2018-12-26
Fix width of fields in imbricated forms

sebj54:
When using translation interface, fields in the imbricated form are overflowing.

**Before: **
image

**After: **
image


benhaynes:
Thanks @sebj54!!! This one was annoying me for a while :)

@rijkvanzanten will check this out as soon as he's back from his holiday.

PR #1304, opened 2019-02-05
Use updateItem for system collections

assmith:
Commit #b27304e introduced a bug that prevents you from updating system tables.  In particular when trying to update a user the app would try to update /item/directus_users instead of /users resulting with the error "Direct Access to System collections are not allowed".

This patch implements the same system table test from the JS-SDK and calls the original api.updateItem for system tables.


benhaynes:
Thanks @assmith!! 😄

PR #1314, opened 2019-02-07
Update project-switcher.vue

dev7ch:
fix selected option for project switcher


benhaynes:
Thanks @dev7ch!!

Question: Is it `project` or `projectName`? I wouldn't be surprised if our Docs are out-of-date... but this example is different: https://docs.directus.io/advanced/app/store.html

@rijkvanzanten ?

PR #1315, opened 2019-02-08
[WIP] Changing project API and user from Admin UI

dev7ch:
Hey there :)

Due this API Change is a great idea and can help you a lot to manage different DB quickly I implemented a basic functionality of the desired feature by allowing login with credentials from Admin UI.

I dunno exactly about your security and login handling and also it`s doing window.location.reload() (ideas how to do it better?)

Hopefully, this can be helpful or at least a basic starting point for this desired feature.

Preview:

ezgif com-resize 1


PostPollux:
Looks nice and useful. I would be happy to see something like this integrated.

dev7ch:
@rijkvanzanten what do you think in general of is implementation?
Is it worth to take a closer look for you?

this is currently more a basic starting point, which can be used as a workaround to switch fast between projects

rijkvanzanten:
I like where this is going. Looks really useful to me. I'm also keeping an eye on https://github.com/directus/directus/issues/2303. We might want to consider making this an admin only thing?

dev7ch:
Yes, I think this is a bit different approach as [directus/directus#2303](https://github.com/directus/directus/issues/2303).
It`s more like a separate "patch for the login screen" i think.

But if you consider merging this I would put a bit more effort on it to make it more smoth and consistent?

dev7ch:
... and further, I asked myself is there some specific method to re-fetch all data from api (instead of `window.location.reload()`)

rijkvanzanten:
> ... and further, I asked myself is there some specific method to re-fetch all data from api (instead of `window.location.reload()`) ?

Ohh the app wasn't really build with this in mind, so I didn't build any `store.reload()` or anything. 

You can maybe use the `resetState` function that's being exported by the store:

dev7ch:
@rijkvanzanten thank you :)
I tried a bit, but it doesn`t seem to work easily out of the box, i guess due `resetStore` will force a logout, but change of the __DirectusConfig__ in the store actions to clean it up a bit.

dev7ch:
you could test it here https://hopeful-bose-50e211.netlify.com with the same logins

dev7ch:
@rijkvanzanten what do you think about this? is this ready and make sense to merge?

dev7ch:
???

benhaynes:
Hey @dev7ch — Rijk got caught up with a last minute integration for a hosting company this week (that we're really excited to announce soon!). But that means he didn't have time to work on core or test the PRs. We're planning on spending Monday on core.

dev7ch:
Hey @ben thanks for the update and dont worry :) best wishes

rijkvanzanten:
This works great @dev7ch 😄 

@benhaynes Could you take a peek at the styling? I'm not too sure about the font-size of the inputs, but you'll most likely have the answer there 😉

dev7ch:
thank you for having a look into it @rijkvanzanten 👍 
@benhaynes please let me know if there are some styling issues, I ll be happy to fix them :)

dev7ch:
Thank you for the detailed review!

> ... since the new login page would auto-select the chosen API based on the project chooser:

Yes, this feature was initially made coz the API chooser wasn`t related to the selection in Admin UI.
I think would be more a feature for SuperUser (another issue, decoupled user role table) to manage multiple installations from one admin.

### Dropdown vs Header

yes, not ideal ... what about moving the lock icon to the bottom of the form above the line?

image

Currently, the "red chooser" is the indicator of a failed login.

### Cancel

... is definitely missing! could be simply done by a page reload. would that be ok?

### SSO
Upps, is havent test it .. i ll have a closer look to it and come back.

### 7.1 Redesign

Yes i think too it`s to much.
Actually, there is still some bug, coz now if you click to a collection when the login is active you get an error.

 maybe an idea would that if the login is visible all other items (Collections, are hidden.

Finally, i think if you don't think 100% this is a feature which brings value to the User Experience, better lets drop this ;)

dev7ch:
@benhaynes do you think it would be better to just fix the login pre-selection instead of implementing this?

benhaynes:
It seems like that would be cleaner, and it's something we have to do anyway for Cloud (this week). I hate to abandon a pull requires though...

Just not sure hope we'll fit all these alternate auth options into that space. 😬

dev7ch:
> Just not sure hope we'll fit all these alternate auth options into that space. 😬
yes, @dev7ch  i can understand your doubts totally.
Due it uses your default login methods i think from the functional side it should just work, it might be more question of UI Design and real UX benefit issue.

Or what do you think about a User profile setting e.g "Allow Login from Admin interface" ?
So that won't be visible by default and an optional addon?

Regarding the Benefit and UX, I think for some circumstances it could be "cleaner" to just stay in the admin interface and change with 2 clicks user and/or remote api.

Further i was thinking about to allow to change the user of the current, selected API as well, which could be handy for admin or managing editors to real test user accounts?

Toughts?

benhaynes:
True — functionally this should work fine since it's the same code.

To me this is strictly a visual change, but doesn't seem any more efficient. Either way the process is:

1. Choose option in Project Chooser
2A. Type in credentials (on login page or sidebar) and click login
2B. Click SSO option (login only?)

_This assumes we fix the login to accept a project query string to auto-set the dropdown._

It might seem cleaner to do in the sidebar, but then we have two different login designs — and I feel like consistency is more important. Also, this would limit how we can change the sidebar in the future since it would need to support the login flow.

Also, I'm concerned about how we would replicate all the other features that will exist on login:

* SSO options
* Forgot Password
* Custom Logo
* Register (maybe not applicable here)
* "Cancel" (only needed for project chooser)

Even if we displayed the login for as an overlay (just like the current login page, but smoking out the page like the modals do) many of these issues are solved and we can use the same view.

Would moving to an overlay be a good middle-ground?

benhaynes:
Also, thank you @dev7ch for taking the time to discuss this. I'm always 100% down to chat through the best way to handle these sorts of things. 😃

syslogic:
> ... and further, I asked myself is there some specific method to re-fetch all data from api (instead of `window.location.reload()`) ?
> 
> https://github.com/directus/app/pull/1315/files#diff-2ed7e13a6b898c156586710de1d131c5R88

[state management](https://vuejs.org/v2/guide/state-management.html) doesn't sound too unlikely.

rijkvanzanten:
@syslogic see my previous comment https://github.com/directus/app/pull/1315#issuecomment-465372314 😉

dev7ch:
@syslogic yes, I think too that would be the way to go to, but I met some circumstances (https://github.com/directus/app/pull/1315#issuecomment-465492734) which let me go straight to a simpler solution (reload).

Further this seems to break a bit the styling and desired UX behavior of directs.

Once the dropdown selection works corresponding with login panel this feature could be omitted.

So i am gonna close this PR and leave this as an working example or a temporary fix until the desired core features is fixed.

syslogic:
@dev7ch just had a look at `resetState()` and resetting the global state to it's initial value would indeed logout the user altogether. think that in order to make this work as intended, one would need one global application state and a collection of states per API account added there, where selecting (or logging in) one such account would propagate it's state to the global state.

PR #1343, opened 2019-02-16
Handle invalid collection name

samvasko:
This PR fixes two minor issues related to collection creation modal dialog

**Stop the indicator from spinning if error occurred**
> Fixes an issue where indicator would not stop spinning even when response was
received.

**Add error message for invalid name**
> Currently this is the only way the API will return error code 4 using this form
is for the user to provide invalid name.


rijkvanzanten:
Thanks! Can you make the branch up to date? It doesn't let me do it on mobile.. I'll merge it right away 💪🏻

benhaynes:
Thank you so much @samvasko!

PR #1344, opened 2019-02-17
Fix double scrollbar when sorting is enabled

samvasko:
Fixes an issue where two scrollbars would appear due to incorrect width of the
table header when a collection has a sortable field.

Example of the bug:
screen shot 2019-02-17 at 14 31 05


benhaynes:
Awesome — such an annoying bug. Thanks! :)

PR #1350, opened 2019-02-20
Allow camelCase as DB table names

dev7ch:
should fix https://github.com/directus/api/issues/783

plz just close is there are some aspects which i haven't considered


benhaynes:
We'll take a look tomorrow! Thanks @dev7ch

rijkvanzanten:
Should we do the same for field names? @dev7ch

dev7ch:
mmh why not ... just keep in mind it could lead into different behaviors depending on OS,
e.g. (afaik)
- windows table or field names => not case sensitive
- unix table or field names => case sensitive
- using InnoDB engine => case sensitive on unix and windows

rijkvanzanten:
hmmmm ok maybe smart to play it safe and let it be as is than 😄

Mohamed-Hamed:
Hello,
I have just downloaded the directus suite to use it for an existing database, and I am still having this issue. Column already exists when I try to change the interface of a camel case column.

benhaynes:
Thanks @Mohamed-Hamed — we're aware of this issue and I believe that @bjgajjar has already fixed it on the `develop` branch.

Mohamed-Hamed:
Ah! Thank you, I thought it was fixed in this pull request, but found that it was reverted buy @dev7ch later on march 22.

PR #1362, opened 2019-02-26
Fix select safari

### LINKED ISSUE #1360: API Selector Broken in Safari
# Bug Report

CSS issue in Safari selector on app editor page (working on login page) affecting ease of activation and display of API selector menu.

Bug observed on live demo & latest docker install.

I believe that this is unrelated to https://github.com/directus/app/issues/805

## Steps to Reproduce

1. In Safari ~v12, macOS, log in to directus (i.e. https://directus.app)
2. Click directly on name of current API instance (i.e. Directus Stable) in API selector in top-left corner
3. Select menu doesn't appear
4. Click above name of current API instance in selector
5. Select menu will appear higher than expected

## Expected Behavior

API selector menu's `<select>` should fill height of its container and open when clicking anywhere in container.

## Actual Behavior

The `<select>` field is using a default height and as such is only clickable from the top of its container down.

Safari doesn't support styling of `<select>` elements out of the box which prevents the height being applied, refer to https://stackoverflow.com/questions/13252835/safari-change-select-height which proposes a range of solutions.

Applying `-webkit-appearance: menulist-button;` to the `<select>` appears to fix the issue, and brings the display of the selector pretty much inline with Chrome.

The API selector on the login page has no such problem, there the `<select>` element's height and position matches that of the label text. Note that applying `-webkit-appearance: menulist-button;` to the login API selector results in a narrower inherited font replacing the system font normally used for the selector - which is a little unnerving :roll_eyes:

## Technical Details

- OS: macOS 10.14.3
- Browsers: Safari v12.0.3, Safari Technology Preview v12.2
- Directus App: v7 (live demo / also observed in latest docker container)

dev7ch:
Closes #1360


benhaynes:
Thanks @dev7ch!!

dev7ch:
thanks for instant merging ;)

FYI as far i know `fix` and `closes` have the same functionality @rijkvanzanten

rijkvanzanten:
Ohh I didn't see you put it in the commit name, my bad 🙂

Issue #1261, opened 2019-01-21
Sort descending


dur41d:
# Feature Request

Make the sort field show items in descending order.

## What problem does this feature solve?

I have blog articles that I would like to show in descending order (latest first). However I would like also to control the sort order manually. That's why I'm using a sort field for my collection. However the admin interface shows the items only in ascending order when using the sort field that means that my latest article will be at the end which is a problem when I have hundreds for articles. 

## How do you think this should be implemented?

The sort icon show allow ascending and descending order just like other fields.

ceruulean:
I think this is already in the app?
![image](https://user-images.githubusercontent.com/41603229/52818446-6fb14300-306c-11e9-82a8-662e6abd00bf.png)

rijkvanzanten:
I think he's referring to the manual sort button in the top left of the listing view @ceruulean. While regular columns can be sorted asc and desc, the manual sort button only supports asc sorting

ceruulean:
Ooo ok, how about clicking on "Sort" field stays in Manual mode?

dur41d:
When choosing the manual sort the items are sorted by the sort field and ascending order only and there is no way to make it sort in descending order.

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

PR #1366, opened 2019-02-28
Manual sort allow desc

ceruulean:
For #1261
Basically if the sort field is visible on the table and already sorted, switching to manual mode will not change the sort view and works like normal

### DIFF
--- a/src/components/table/table.vue
+++ b/src/components/table/table.vue
@@ -341,13 +341,12 @@ export default {
     },
     updateSort(field, direction) {
 
-      var newSortVal = {
+      const newSortVal = {
         field,
         asc: field === this.sortVal.field ? !this.sortVal.asc : "ASC"
       };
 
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @ceruulean — @rijkvanzanten will take a look as soon as possible!

rijkvanzanten:
While it works as advertised, I think the actual usability doesn't feel logical to use. If I sort by any column, then change the order manually, the sort is only preserved for that specific column. 

Instead, I think we should mimic the sort behavior of the regular columns in the manual sort column as well. EG first click will set ASC, second click will make it DESC and the third click will disable it again. (And still have it sort only by the sort column)

rijkvanzanten:
I really appreciate the effort @ceruulean! But unfortunately this is not the solution to https://github.com/directus/app/issues/1261

PR #1368, opened 2019-03-01
/deep/ is deprecated

okayawright:
See https://developers.google.com/web/updates/2017/10/remove-shadow-piercing


rijkvanzanten:
Does Sass accurately parse this yet?

We're not using webcomponents, so the deprecation in that spec doesn't directly apply to the Vue based codebase. We're relying on Vue's scoped CSS deep selectors: https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors. As per those docs:

> Some pre-processors, such as Sass, may not be able to parse >>> properly. In those cases you can use the /deep/ combinator instead - it's an alias for >>> and works exactly the same.

okayawright:
> Does Sass accurately parse this yet?
> 
> We're not using webcomponents, so the deprecation in that spec doesn't directly apply to the Vue based codebase. We're relying on Vue's scoped CSS deep selectors: https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors. As per those docs:
> 
> > Some pre-processors, such as Sass, may not be able to parse >>> properly. In those cases you can use the /deep/ combinator instead - it's an alias for >>> and works exactly the same.

Hello, thanks for the review,

apparently the latest Sass 1.7 understands it as far as I can tell.

PR #1380, opened 2019-03-09
Fix drag and drop not working with v-upload accept prop

binary-koan:
Based on feedback on https://github.com/directus/api/pull/747

TIL file inputs ignore the `accept` attribute when you drag and drop files (https://stackoverflow.com/questions/49001554/file-field-accept-attribute-not-working-properly-when-drag-and-drop-file-into), so that needs to be handled separately in the `v-upload` component.

This makes the same assumption as the `file` and `files` interfaces, that `accept` is a comma-separated list of specific MIME types, not the full syntax supported by HTML.

### DIFF
--- a/src/components/upload.vue
+++ b/src/components/upload.vue
@@ -123,6 +123,15 @@ export default {
       embed: false
     };
   },
+  computed: {
+    acceptTypesList() {
+      if (this.accept) {
+        return this.accept.trim().split(/\s*,\s*/)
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
> TIL file inputs ignore the accept attribute when you drag and drop files 

Some browser APIs keep surprising me (don't get me started on `datetime-local`)

PR #1383, opened 2019-03-12
Enhancement / <v-upload/> component: add preview feature for embedded files

alex-gorovyi:
Enhancement / **v-upload** component: added preview for embedded resources after successful creation of an item.


benhaynes:
Thanks @alex-gorovyi !!

rijkvanzanten:
What happens if the user pastes a YouTube or other non-image link?

alex-gorovyi:
> What happens if the user pastes a YouTube or other non-image link?
Good question, @rijkvanzanten .

My changes just use existing preview functionality that UPLOAD file uses.
I just tried it with YouTube link and this is how it looks:



I am happy to enhance that part as well, if someone could point mw to the right direction and suggest how the preview for YouTube should look like.

rijkvanzanten:
Oh this is actually perfect 🙂 

I was wondering if it would break down completely 😉

PR #1389, opened 2019-03-14
Configurable auto sign out from general settings

bjgajjar:
Currently, in the general setting, auto sign out is already there. We just need to update the variable to make it work.


benhaynes:
Thanks Binal!! Your first merged PR on Directus! 😄

PR #1403, opened 2019-03-18
devDependencies updated & "Help & Docs" button fixed. 

syslogic:


rijkvanzanten:
Thanks @syslogic! 😄

syslogic:
@rijkvanzanten Once I've had the "Help & Docs" link working, it made me wonder if that nav-menu is even a suitable place for such an external link. It's the second item from top, which is a quite prominent position - and it's not alike as if it would display the documentation in a windowed iframe - or as contextual help.

rijkvanzanten:
I see what you're saying, but I do believe it's important to have a link to help and docs available to every user

benhaynes:
I don't think it's that distracting. We need to have it somewhere, and that seems like the most non-intrusive.

syslogic:
Don't think it's distracting, but it still feels a bit unnatural (UX wise), when a SPA opens a link in new tab (on-site vs. off-site content). I was also wondered if there is a better location for the button, but as the UI currently is organized, it would rather interfere when putting it eg. somewhere top-right. I there is a distraction, it's these 3 menu dots, looking alike a drag-handle, which do not seem to serve the least purpose.

benhaynes:
True — perhaps the "solution" is to have the docs rendered internally within a modal. Since our Docs are all in a markdown repo, we could even add this to our build process and have the docs combined into the codebase... this has the added benefit of showing the docs of the version you're actually using (instead of showing the latest docs for an older install).

benhaynes:
Also, we'll update the "three dots" icon on the user menu to an arrow that has clearer meaning.

Issue #125, opened 2018-04-11
Navigation Override Option to Customize Sidebar


rijkvanzanten:

rijkvanzanten:
DIdn't know I would say this, but it should be a little more WordPress menu creator style

rijkvanzanten:
Ref https://github.com/directus/app/issues/456

benhaynes:
I'm fine with JSOn for now, but here are the things we want to support:

* Shorthand for keeping dynamic sections (Collections, Pages, etc)
* Linking to other internal pages (such as specific Item Detail)
* Grouping (uses standard sidebar headers/labels)
* External/custom links
* Maaaaaybe nesting (I'm not sure about this one)

So, the example below has custom collections listed in a custom order, then has a _second_ group of other collections. There are also links to individual Item Details and external links. The Bookmarks have a `true` value, so those load the default values... but the order has been reversed (bookmarks _before_ extensions).

```
{
    "Collections": {
        "Z. Collection Name": "/collections/collection-name",
        "X. Collection Name": "/collections/collection-name",
        "Y. Collection Name": "/collections/collection-name",
        "Specific Item Name": "/collections/collection-name/123",
        "External Link": "https://example.com"
    },
    "Another Group": {
        "A. Collection Name": "/collections/collection-name",
        "B. Collection Name": "/collections/collection-name",
        "C. Collection Name": "/collections/collection-name"
    },
    "Bookmarks": true,
    "Extensions": true
}
```

rijkvanzanten:
Lets use `$bookmarks` instead of "Bookmarks", otherwise there's no way to tell what are the system groups vs the user group names, and lets use it as value instead of title, so you can override the group names as well 🙂 

```json
{
    "Collections": {
        "Z. Collection Name": "/collections/collection-name",
        "X. Collection Name": "/collections/collection-name",
        "Y. Collection Name": "/collections/collection-name",
        "Specific Item Name": "/collections/collection-name/123",
        "External Link": "https://example.com"
    },
    "Another Group": {
        "A. Collection Name": "/collections/collection-name",
        "B. Collection Name": "/collections/collection-name",
        "C. Collection Name": "/collections/collection-name"
    },
    "Favorites": "$bookmarks",
    "X-tension": "$extensions"
}
```

andreekeberg:
[...]

Issue #721, opened 2018-09-07
Warn users if their account is logged in at another location


benhaynes:
#### :rotating_light: MIGRATED FROM REQUESTS.GETDIRECTUS.COM :rotating_light: 

# Warn users if their account is logged in at another location
### :+1: = 2
*Created 1 year ago by @benhaynes*
 

-----
### @benhaynes – (1 year ago)
It would be useful to know if you are logged in elsewhere (for security reasons). We should warn users when they log in:

Your user is still logged in at another location: (000.000.000.000)

PR #1405, opened 2019-03-18
Add nav override option to directus_roles

rijkvanzanten:
Closes https://github.com/directus/app/issues/125

This PR adds a new nav override field to the roles pane 

image

This field allows the admin user to add a JSON object that will override the main navigation menu.

image


src/store/actions.js
```diff
@@ -52,15 +52,15 @@ export function getCurrentUser({ commit }) {
...
+        "roles.*.*"
       ]
     })
     .then(res => res.data)
     .then(userInfo => {
       return {
         ...userInfo,
         roles: userInfo.roles.map(obj => obj.role),
-        admin: userInfo.roles.map(obj => obj.role).includes(1)
+        admin: !!userInfo.roles.map(obj => obj.role).find(obj => obj.id === 1)
```
YvanDaSilva: Hello,
I know this merge was done a while ago. However, with the latest API / APP from master. 

The userInfo does contains roles. But there is no "sub role" object with an id of 1.
But instead it userInfo.roles contains an array as follows.

```
roles: [{
id: 3,
role: 1,
user: {id: 3,....}
}]
```

Meaning the condition on line 63 will always be false, and even as an admin, the interface believes it is not an admin.

YvanDaSilva: @rijkvanzanten Let me know if you need more information.
The DB schema is up to date and I have seen your changes related to #721 on the API side and nothing seems wrong.

rijkvanzanten: Is this tied to a specific bug? Please open a new issue if that's the case, as this could've been changed in the meanwhile 🙂

PR #1410, opened 2019-03-19
pgsql 10 initial support

okayawright:
Initial support for PostgreSQL 10.

Can be potentially merged in a separate branch for PostgreSQL beta testers to play with.

Grouped PRs made available for the API, APP and the sdk-js at the same time, but for each repository independently.

Known limitations so far:
 - Error if one tries to transform the type of a primary key column, PostgreSQL cannot change an identity column into something else afterward

P.S. didn't properly perform any regression tests on MySQL


src/components/permissions/permissions.vue
```diff
@@ -194,7 +194,8 @@ export default {
       border-top: 1px solid var(--lightest-gray);
     }
   }
-  /deep/ .cell {
+
+  >>> .cell {
```
rijkvanzanten: We actually had to revert this change you made earlier from `/deep/` to `>>>`, as vue-cli uses an older version of sass that doesn't support this yet...

```suggestion
  /deep/ .cell {
```

src/lang/locales/it-IT.js
```diff
@@ -162,6 +162,27 @@ export default {
         "Una data e ora. Permette `1970-01-01 00: 00: 01` UTC a` 2038-01-19 03: 14: 07` UTC.",
       TIME: "Un tempo. Permette `-838: 59: 59` a` 838: 59: 59`.",
       YEAR: "Un anno. Permette di `1901` a` 2155` o `0000`"
+    },
+    pgsql: {
+      CHARACTER: "<placeholder>",
```
rijkvanzanten: We should remove these `<placeholder>` values. If a value doesn't exist, the application will fallback to the English string. Now, the user will see `<placeholder>` in the application

benhaynes:
Thank you @okayawright!! We'l continue to monitor/review this! :)

src/lang/locales/it-IT.js
okayawright: Removed all placeholder l10n text for pgsql

benhaynes:
Hey @okayawright — I know this initial push covers a lot! I'd love to announce this new branch to the community to get some beta testers, but want to make sure that it's in a decent place first. Would you say that this is ready to have users start reviewing? Are there any specific Docs that we should have to outline differences for admins or is everything essentially the same?

PR #1414, opened 2019-03-21
Login form & file upload error messages

syslogic:
Webkit browsers do not emit a change event on auto-fill and so the auto-filled form fields will not have class `has-value`, as they're supposed to have (placeholder & content were visually overlapping on fields `email` & `password`). Firefox does not automatically insert the credentials into the form fields; instead one has to select them first, therefore it is not an issue, because this triggers a change event. Just noticed, that on Firefox the labels are slightly overlapping the form fields, when clicking submit.

Also displaying the actual `error.message` on XHR file upload, instead of the generic `something_went_wrong_body` message.


src/components/upload.vue
```diff
@@ -244,7 +244,7 @@ export default {
         .catch(error => {
           this.files[id].error = error;
           this.$events.emit("error", {
-            notify: this.$t("something_went_wrong_body"),
+            notify: (typeof(error.message)!=='undefined' && error.message!=='' ? error.message : this.$t("something_went_wrong_body")),
```
rijkvanzanten: Can we change this to a "regular" if statement above?

```js
let message;

if (error.message) {
  message = error.message;
} else {
  message = this.$t("something_went_wrong_body");
}
```

I know it's more verbose, but it's way easier to read 🙂

src/lang/index.js
```diff
@@ -37,9 +37,10 @@ function setI18nLanguage(lang) {
  * @returns {Promise<string>} The language that was passed
  */
 export function loadLanguageAsync(lang) {
+  if (typeof(lang) === "undefined") {return;}
   if (i18n.locale !== lang) {
     if (!loadedLanguages.includes(lang)) {
-      return import(/* webpackChunkName: "lang-[request]" */ `@/lang/locales/${lang}`).then(
+      return import(`@/lang/locales/${lang}`).then(
```
rijkvanzanten: What does this change do? Isn't the webpack comment needed for the lazy loading with the correct file name?

src/components/upload.vue
syslogic: have reverted this to the default error message.

src/lang/index.js
syslogic: also reverted this change, while leaving a comment about it's purpose.

syslogic:
just not sure why `typeof lang === "undefined"` fails on SemaphoreCI; this evaluates in the console.

rijkvanzanten:
Semaphore updated to a new major version lately, I've seen a lot of false negatives on the CI since :/

rijkvanzanten:
What is that core-js dependency for exactly @syslogic ?

syslogic:
> Semaphore updated to a new major version lately, I've seen a lot of false negatives on the CI since :/

this might come from the [cancellation strategy](https://semaphoreci.com/docs/set-queued-builds-cancellation-strategy.html)... because it happened when I've probably pushed further commits too quick.

and the `core-js` is being demanded by babel. without it, the build complains several times about:

    "With `useBuiltIns` option, required direct setting of `corejs` option".

rijkvanzanten:
The new `handleAutofill` method actually fires every single time, not just on autofill, causing the labels to be in the `has-value` state constantly. I've instead added an additional css selector for `:-webkit-autofill` that will cause the same behavior as the `has-value` class.

syslogic:
have to admit, that I haven't tested it with the auto-fill password removed from the browser, nor when removing the auto-filled password from the form... the `JS` code does indeed not remove the set class again.

PR #1428, opened 2019-03-27
Consolidate references to the hosting path.

skeemer:


vue.config.js
```diff
@@ -7,5 +7,5 @@ module.exports = {
     appleMobileWebAppCapable: "yes",
     appleMobileWebAppStatusBarStyle: "black"
   },
-  publicPath: process.env.NODE_ENV === "production" ? "" : "/"
+  publicPath: process.env.NODE_ENV === "production" ? "/admin/" : "/admin/"
```
rijkvanzanten: This doesn't feel right to me. `/admin` is only used in the combined build, but isn't necessarily the standard for the application itself. What does this change fix exactly?

skeemer: Ahh, I actually meant to have "/admin/" : "/". The first being because that is the the combined build and a good example of how to use it.

skeemer: Actually, leaving it empty wouldn't work with the optimizations in index.html. At the minimum we would need to do "/" : "/". I can add that if you prefer.

rijkvanzanten:
Leaving vue.config.js publicPath empty `""` means it will [build the static assets using relative paths](https://cli.vuejs.org/config/#publicpath). Adding `/` will break running the application from any folder.

I just tried building the app, and can confirm the app won't work unless the static assets are in an `/admin` folder

rijkvanzanten:
> Leaving that blank as "" is just asking for configuration trouble.

What do you mean exactly @skeemer? Currently, downloading the build of the application will work as expected in any subfolder, and running the app locally to work on it works fine too. What configuration trouble are you trying to solve?

skeemer:
My main concern was using 'history' routing and I had to change 3 or 4 files to make it work. That's not a good experience.

rijkvanzanten:
Oh! In what way didn't it work exactly? Setting the `routerMode` option in `config.js` to `history` does the trick for me

skeemer:
I'll have to go back an do some more checking, but all of the "./filename paths" in index.html failed completely.

rijkvanzanten:
Hey @skeemer! Did you have a chance to verify this?

rijkvanzanten:
I'm closing this as I can't reproduce the original issue and there hasn't been any activity in a couple months.

PR #1436, opened 2019-03-30
Add global bookmark functionality

honzabilek4:
Hey, 
I've implemented the global bookmark functionality on frontend. See a screenshot below. I've added a simple toggle, where you can specify a global bookmark (false by default).
I also made a few assumptions on the way:
- Only admin can create this type of bookmark
- Other users than cannot see the toggle

TODO: I think there should be some permission check on the API for the following:
- Regular users shouldn't be allowed to delete a global bookmark. (I could possibly hide the delete button for now)
- Regular users shouldn't be allowed to create it either (less important)
- Missing translations of the description to other languages

Screenshot 2019-03-30 at 23 41 17


rijkvanzanten:
> (I could possibly hide the delete button for now)

This sounds good to me 👍 


Should we also add an additional toggle / dropdown that allows admins to set a bookmark on a Role level?

benhaynes:
Yeah, we definitely need to hide that delete button except for Admins. 

If we did want to add support for Roles, I think we could use a single dropdown for admins:

* Personal: Just Me (default)
* Global: All Users
* Role: Administrator
* Role: Manager
* Role: Intern

honzabilek4:
Is it possible now on the api to assign bookmark to a role instead of user or null?

rijkvanzanten:
> Is it possible now on the api to assign bookmark to a role instead of user or null?

By setting the user to `null`, but leaving the role it should be assigned to the role and not the user. Also, the API itself doesn't care about any of this. It's just the application reading and using the rows.

benhaynes:
Hey @honzabilek4 — just curious how things are going with this! Any updates on what's left to do? Or, is it ready to review/merge?

honzabilek4:
Hi, I'm on it. It's almost ready for review, so expect a PR later today. 😎

honzabilek4:
Should be ready for review now. Can you please take a look and test it? (Works on my machine 🤓😅)
I've added a dropdown to select what type of bookmark user will create. (Global, Personal, Role specific). This dropdown is visible only for admins now. Regular users can create only personal bookmarks.

rijkvanzanten:
The only thing I just thought of is that you theoretically can have a global bookmark that points to a place you don't have access to 🤔 

(not an issue with this PR per se, but more of a thought problem in general)

rijkvanzanten:
_Works on my Machine™_ 😄

PR #1446, opened 2019-04-03
Alias fields should be used on the whole GUI to filter data

okayawright:
This is a generalization of #1433 to other areas of the APP

One should not offer to filter data through the APP with 'alias' fields
Now both Files and Users cannot be filtered with invalid fields

It is related to the PR **Alias fields should not be used on the whole GUI to sort data** on the API side


rijkvanzanten:
Thanks @okayawright! This PR doesn't require the API one to be merged to function right?

okayawright:
No, indeed, you can separate them.

PR #1451, opened 2019-04-04
New Feature: Suggest options for relation templates

shealavington:
Adds a new button next to the relation `template` field.

Allows you to select fields from the related column, saves you from needing to remember valid column names.

:) 

image


benhaynes:
Love this — these little options make the interface so much more intuitive! :)

rijkvanzanten:
I have one small problem with the implementation though..........

This relies on the name `template` in the interface options to render this specific interface tool. This has a couple problems: 
1) The interface option may or may not be called `template` (vs `display_template`)
2) Other interfaces can use an option called `template` for another purpose
3) Users can create their own interfaces with options that are names whatever. Therefore we should consider interface options user generated content which we can't rely upon.

Instead, I think this should be a separate interface that the o2m / m2m interfaces then use 🙂

rijkvanzanten:
And by having it as an interface, we might also be able to use it for the mirrored field option of the slug :sligh

shealavington:
Updated!

M2M (Template & CSV): https://i.gyazo.com/f80ad723206546085a2eaf770592f013.gif
M2O (Template): https://i.gyazo.com/d6329b4ab2a8a2329f37d5f3c7fe979c.gif
Slug: https://i.gyazo.com/a4739de378268c8b8f7aaf032e4b1d90.gif

---

The interface (text-input-suggest) configuration below, allows for:
* "Replace String": If true, clicking an item will replace the entire string, if false, then append item it to the end of the string ( Slug )
* "Suggest Current Collection": Can select fields from the 'current' collection ( Slug will mirror another field in this collection )
* "Prefixing and Suffixing": It will prefix and suffix items. So you can add templates and CSV ( `{{first_name}} {{last_name}}` ) ( `first_name, last_name` )
* "Single-Collection" suggestions (`first_name`)
* "Multi-Collection" suggestions (`members.first_name`)

```
var options = {
  "replaceString": false,
  "suggestCurrentCollection": false,
  "prefix": "{{",
  "prefixCollectionName": true
  "suffix": "}}",
}
```

:smile:

rijkvanzanten:
> ( 03:35am... I did say that I usually go until 4am

Evident by the classic past- 2am commits

image

😄

benhaynes:
Hey @rijkvanzanten — let's plan on reviewing this tomorrow (unless you have more feedback). Sorry for the delay @shealavington!

This would also be a good interface to use in that `alias` interface we were discussing in #general (that lets you create a `{{first_name}} {{last_name}}` concat sort of field).

rijkvanzanten:
@shealavington How does it work exactly? I noticed it's trying to read from `this.values` which throws an error for me

```js
this.values.suggestCollectionFields.forEach(collectionName => {
  collectionFields[collectionName] = this.$store.state.collections[collectionName];
});
```

```
TypeError: Cannot read property 'forEach' of undefined
```

This happened when trying to setup a M2M interface. Do I need to set anything up?

shealavington:
I think it should be this.options, sorry trying to review it on my phone 😅 off the top of my head, I can't remember haha.

rijkvanzanten:
The interface has no way of knowing what the "collection" is the user is referring to in the relationship. Therefore it can't render the list based on the fields currently available. I am a big fan of the functionality itself, but I'm afraid it can't be done in the current architecture of the field-setup flow. Let's revisit this at a time where we can provide some internal API for these fields.

PR #1470, opened 2019-04-10
Don't make 'alias' fields available on the APP to sort data

okayawright:
Globally hide on the GUI references to 'alias' fields in comboboxes used to sort data, because anyway the API rejects such fields as sort columns


benhaynes:
Thank you for this @okayawright! Rijk might have time to review tomorrow, otherwise we'll check this out first thing Friday.

PR #1474, opened 2019-04-10
color interface: fix errors and improve code base

dev7ch:
i like the style of your color interface a lot.

but i was not happy with the console logs so i in this PS is:
- fix the primitive value errors 
- standardize model Alpha Value
- a little bit DRY improvements


benhaynes:
Awesome! Thanks @dev7ch!! 😄

PR #1486, opened 2019-04-13
Update .gitignore

dev7ch:


rijkvanzanten:
I always thought you were supposed to leave the lock files in the repo 🤷🏻‍♂️

dev7ch:
Yes but if you have both yarn and package lock it gives you a warning or could lead differences

dev7ch:
e.g if you use yarn locally and switch between branches u have to remove the yarn.lock to keep the upstreams clean

Issue #1496, opened 2019-04-15
Cannot soft delete a collection anymore if the "status" field has been deleted


okayawright:
If you try to delete a collection that has lost its **status** field then an error will occur

# Bug Report

## Steps to Reproduce

1. Create a collection with a **primary key** and a **status** field
2. Modify the collection and remove the **status** field
3. Do not refresh or reload the application and try to delete the collection

## Expected Behavior

The collection is deleted

## Actual Behavior

An error occurs

## Other Context & Screenshots

Apparently, the application tries to *soft delete* (change the state of the collection to hide it from sight) thinking the **status** is still present but fails

## Technical Details
Latest master

rijkvanzanten:
Closed in #1497

PR #1497, opened 2019-04-15
BUG #1496

okayawright:
Tries to fix issue #1496 but I'm not sure the solution is clean enough


rijkvanzanten:
LGTM!

I'm thinking to change the structure of the store a little bit, so it saves permissions "raw" like they come from the database. Then I'll add a `getter` that allows you to request the permissions for a given collection and field. This getter can then in turn check if the status field even exists before returning the wrong permissions. Anyway, that will take a little while to properly implement, so this is perfect for the time being

PR #1498, opened 2019-04-15
FEAT restrict actions on the APP for PK fields

okayawright:
I have made a change on the APP that prevents a user from both removing a *primary-key* column and modifying key elements from it (more or less you can only change its comment now); it's a breaking change for Directus and I'm not sure that it would be beneficial for most users.
My users believe that any kind of modification on a *primary key* is deemed too risky and prone to errors so they asked me to remove that feature.

This PR provides the following features:
- You cannot remove a *primary-key* field in a collection from the APP

image
- Most options for a *primary-key* are disabled during edition from the APP

image
- For non-*primary-key* fields, the *primary-key* interface is not available during edition from the APP

image
- For "primary-key" fields, only the *primary-key* interface is available during edition from the APP

image

- Bonus: if you click on the currently used interface icon for a field during edition, nothing happens instead of saving any change and closing the popup


benhaynes:
Hey @okayawright — thanks for this! I can certainly see the benefit of not allowing changes to the PK after creation. Especially since so much of Directus relies on that key (and its datatype).

My only concern is that we automatically create the PK when creating a collection, and that modal doesn't have any options (yet) for configuring it. That means that when you create a collection inside Directus you're _always_ editing it if you want to make adjustments (before you've actually started adding content to it).

For example, this workflow would mean that you could never create a PK with a String (eg: uuid) datatype. So perhaps we need to add some PK setup to the "Create Collection" modal? I imagine the important info is:

* PK Field Name
* PK Field Type
* PK Datatype (maybe this can be set from the field type?)

Still, I like your solution. But maybe this restriction should be toggled depending on your development mode (eg: `development` vs `production`) so that Directus still functions as a proper "database client" for admins crafting new schemas... but then locks when you're delivering content.

@rijkvanzanten thoughts?

okayawright:
Spot on arguments. The problem with this PR is that it removes a lot of freedom for users indeed.

On a side note, PostgreSQL cannot change the type of a PK afterwards once created (unlike MySQL apparently), it's a limitation of the engine: you must first drop the key and then recreate it, which is something that nobody would be willing to perform in production for obvious reasons.

rijkvanzanten:
I agree with both. I think changing the primary key shouldn't be possible _after items have been created_. Like @benhaynes said, deleting and re-creating the primary key is the only way to create a string based primary key at the moment... If we can add settings for the primary key interface to the create-a-collection modal so the user is able to choose what kind of primary key to use, we can safely disable editing of the primary key afterwards.

benhaynes:
I think we disabled deleting the PK as of now... but perhaps we can allow for deleting it, but can _lock_ (or not "manage") that collection until it has a PK again?

rijkvanzanten:
I'm closing this PR as the implementation completely prevents using any non-numeric fields that aren't called ID. We should re-imagine the create-collection flow to allow the user to setup the primary key during collection creation, at which point we can safely disable any editing of the field afterwards

PR #1500, opened 2019-04-15
Add manual sorting to M2M Interface

### LINKED ISSUE #1388: M2M Interface needs drag/drop sort
This was in v6 and I'm really surprised we missed it. If the junction collection has a `sort` field then we need to allow for dragging and dropping items into a manual new order.

brocgailit:
Resolves #1388 

image


benhaynes:
Whoa — thank you @brocgailit! I know Rijk was saying this was a complicated one!

We'll review this asap (most likely Thursday or Friday when @rijkvanzanten is online next).

kieranwhite1994:
Hey is there an update on this one - noticed you have a fix and merging is still to occur?

benhaynes:
Our team has been super busy the last few weeks, but we'll try to review it asap! 

In the meantime, you should be able to pull this PR and test it to see if it solves the issue for you (and let us know here!). :)

src/interfaces/many-to-many/input.vue
```diff
@@ -395,7 +450,7 @@ export default {
               });
             }
           }
-
+          console.log("delete payload", newValue);
```
rijkvanzanten: Found one! 

```suggestion

```

src/interfaces/many-to-many/input.vue
```diff
@@ -10,6 +10,23 @@
...
+              class="manual-sort cell"
+              :class="{ active: manualSorting }"
+            >
+              <button
+                class="no-wrap"
+                type="button"
+                v-tooltip="$t('enable_manual_sorting')"
+                @click="startManualSorting"
+              >
+                <i class="material-icons">sort</i>
```
rijkvanzanten: Could you replace these with `<v-icon` components?

rijkvanzanten:
Another question, icons aside, the sorting itself doesn't seem to work reliably:

sort

brocgailit:
@rijkvanzanten That seems to be an issue related to the getFieldQuery inclusions.  When reverting back to `.*.*.*` saving works swimingly...

image

brocgailit:
@rijkvanzanten I haven't had much time to dig around into the PHP API end of things, but I think the issue may reside on the receiving end since the payload seems to be okay from the app standpoint (when including as `.*.*.*`).  Possible circular references?

ThomasKliszowski:
I had a look on this this week and I faced the same issue.
The problem seems to be in the data received from the API and used by the app. In this data, each join collection element reference the main collection element, and each of them reference all the join collection elements.
This script only update the values on depth 1, but the app sends the whole payload with the depth repeating the unchanged join elements.
The API seems to apply depth 1 updates and override them each time it encounters the repeated elements.

I guess you guys know better what is the good way to resolve this, since it seems to be a structural issue between the API and the App.

I hope I was clear enough to help you 😊.

brocgailit:
Considering the nature of the M2M and the undoubted circular referencing that can ensue, doesn't it make sense to break the M2M fields into their own edit/save context instead of the minimal inclusion and saving a giant payload?  Not only does it resolve this issue, but it allows for deep editing of nested M2Ms.

ThomasKliszowski:
I am not so sure that the problem is that actually.
Each join collection record (M2M) references all those M2M records through the main collection reference.
Even if you split the context to only the join collection records, this reference still exists, or maybe I missed something?

brocgailit:
We know a few things here:

1. Including up to `.*.*` works.
2. Including up to `.*.*.*` does not work.
3. Limiting to `.*.*` only make sense if its split into a separate editing context since the nested fields are not included.
4. Limiting the inclusion to _any_ extent is a major wall when dealing with nested M2M fields since they aren't populated after a certain point which is a nightmare in the modal/portal path.  As it stands now, its not possible to edit a M2M from within another M2M inside the modal because of the somewhat arbitrary inclusion limits.

Seems to me, including up to `.*.*` and then creating a new context and saving changes to the nested collection instead of the collection entry point would solve most of this without having to consider the API/App relationship and also allow for nested editing of M2Ms.

brocgailit:
Inclusion limits → #1730

rijkvanzanten:
I reworked the O2M in a way where it only stages the IDs of the related collections, instead of full copies of the related item. I'm gonna work on refactoring the M2M to do the same thing, after which we should be able to work in the changes you made in this PR and get it to work 👍

rijkvanzanten:
I'm closing this in favor of #1774. I've worked in the changes of this PR into that one 🙂 Thanks @brocgailit!

PR #1501, opened 2019-04-15
Update edit layout

rijkvanzanten:
It now supports three distinct sizes: half, full, and fill (page width). It also tweaks the label and note sizes.

Note: I've also tweaked the prettier config slightly to make the HTML more readable, that's what's causing the million updates


benhaynes:
@rijkvanzanten — let's add these (four) size options to the Docs so we can actually call this done! 😄

Issue #773, opened 2018-09-08
Grids, Sections & Regions


benhaynes:
#### :rotating_light: MIGRATED FROM REQUESTS.GETDIRECTUS.COM :rotating_light: 

# Grids, Sections & Regions
### :+1: = 1
*Created 9 months ago by @marcobiedermann*
 

-----
### @marcobiedermann – :+1: (9 months ago)
It would be super great it there would be an option to create layouts / grids / sections to nest modules inside. Take a look at Cockpit for reference: https://twitter.com/getcockpit/status/898848619020914688

rijkvanzanten:
This is a duplicate of https://github.com/directus/app/issues/506, https://github.com/directus/app/issues/759, and https://github.com/directus/app/issues/763.

rijkvanzanten:
Groups can be nested as well, so we could make a tab-group that works with nested tab-groups if you know what i mean

```
tabs
  tab
    title
    content
  tab
    color
    map
```

The outer most tab would control the visibility of the inner tabs, which in turn group the actual interfaces

benhaynes:
Beautiful.

So this would be a `Tag Group` interface? And it would be self-aware if it is the parent (tab container) or a sibling (tab)?

What a great/easy solution...

rijkvanzanten:
AFAIK every component can read info on their parent and it's children components. That would make this possible as the parent tab could check for the existence of tab components in it's children and the children can check if the parent is a tab!

benhaynes:
Perfect. So this is just a single interface!

benhaynes:
This is the more robust ticket, so I'm closing the other one. Here is the relevant info from that ticket:

```
1-1-1-1
1-1 - 2
2  -  2
4     -
FUUUUUUUUUUULL (0?)```

benhaynes:
Maybe this is obvious, but we should have this grid system be fully responsive. That means we'll likely want to wrap the layout in half at a certain point: 4->2 and 2->1

alrick:
@rijkvanzanten I see some fields are grouped at the demo (multiple fields on the same line).
How to do this ? can't find anything in the documentation.

rijkvanzanten:
@alrick That's an initial test of this using the `width` column in the DB. While it technically works, we're not too happy with it yet, as it doesn't provide enough flexibility in certain use cases.

You can set the width column in directus_fields to either 1, 2, 3, or 4. It will flex align the items on either 1/4, 2/4, etc

alrick:
[...]

PR #1502, opened 2019-04-15
Add field width setup dropdown to field options

rijkvanzanten:
Allows the user to specify how wide an interface should be displayed on the page. 

image

image

Closes https://github.com/directus/app/issues/773


Nitwel:
Im not sure in which browser you tested this but in Chrome and Firefox it does not work, the boxes are the right size but they dont wrap arround.
grafik

benhaynes:
Thanks @Nitwel — I'm still seeing a few issues with the new form layout, but the fields are wrapping correctly for me. Can you confirm that this is good for you now?

Nitwel:
Seems fixed.

Issue #1229, opened 2019-01-03
Make File Preview interface work outside of directus_files


creazy231:
# Bug Report

## Steps to Reproduce

1. Go to `collections` and create an new entry using the `File Preview` component.
2. Try to add a new article (in my case) and open the console to view the error.

## Expected Behavior
editable file preview

## Actual Behavior
nothing - can't even upload an image

## Technical Details

- Device: Desktop
- OS: MacOS
- Web Server: MAMPP
- PHP Version: 7.2.10
- Database: MySQL
- Install Method: master manually config

FRSgit:
Hey, is anybody working on this?

rijkvanzanten:
Not currently @FRSgit. You wanna take a swing at it? 🙂

FRSgit:
Thinking about that since having image editor in place would be nice for my use case. Do you have any clues for me where to start?

rijkvanzanten:
This is where the actual Vue component is that builds up that file preview interface: https://github.com/directus/app/blob/master/src/interfaces/file-preview/input.vue

Right now, it's relying on a couple proprietary names / fields that only work in directus_files. For example, it [reads the values of the other fields on the page](https://github.com/directus/app/blob/master/src/interfaces/file-preview/input.vue#L144) to extract what kind of image it is, those fields only exist on the detail pages of directus_files.

I think we should be able to extract most of the information it's currently pulling from the other fields from the value (`this.value`) of the file itself which in turn should decouple it from the directus_files collection

benhaynes:
We can also add interface options to allow the user to set/map any other needed fields. But I prefer Rijk's method of getting it from the main value since that's more "set it and forget it".

FRSgit:
**Sorry for the long comment!**

Okay, I investigated the issue a little bit and have few conclusions. 
First of all mine use case was indeed (as in issue #1425) something which should be handled by `file` interface - there is need for having file uploader which allows user to add image editor functionality.
But then, the question arises - what exactly is `file-preview` interface in the first place and if it's not just a `file` with `value` set and e.g. option `displayName` set to `preview`.
Or (other way around) maybe it's separated **extension** of `file` interface, which inherits some of the logic, or just uses `file` component internally (for handling initial situation - when image is missing)?

[...]

PR #1509, opened 2019-04-18
feat: make file-preview usable outside of File Library

FRSgit:
create moreOptionsButton component for better managment of optional functionalities popovers
make file-preview resubable component
include file-preview as a part of file interface
unify file-preview's toolbar styling with rest of interfaces
add label to aspect ratio select
change crooper.js viewMode to overcome problems background adding at photo's edges
add File Library's response structure (whole values object being a photo) support to file interface
add imageEditor option to file interface configuration
add missing translation to file interface configuration
bump up file interface version

fixes issue #1229


FRSgit:
There is still one big problem - thumbnails should be regenerated after each file's crop. But unfortunately I cannot find any `reset`/`force_regenerate`/`remove` method in php Thumbnailer. Am I missing something?

As I see, unfortunately my work got between some big layout changes (and not only?) in Directus. I tried to unify file-preview with old layout and hopefully it's still okay now - not sure though. Also, I don't know why after merging latest commits from master branch (so, commits from last week) I started getting error in file directory - something broke during development:
image

And a last thing - also after last merge deleting image stopped working for file directory (from custom collection it works like a charm). But before fixing anything - is "removing image" functionality needed in file directory context? I mean, if user wants to add new photo, then he should add it from file directory list, not change photo inside of the file entity. On file detail page on editing should be possible.

benhaynes:
Ahh, yeah — some big design changes this week in the App. Rijk will be online tomorrow and can review what's up with the File interface.

And if the thumbnailer doesn't have a parameter to force refresh the image then we should definitely add that! This issue would also happen when replacing a file too.

FRSgit:
Will be continued under #1655

PR #1511, opened 2019-04-18
Improved I18n coverage

okayawright:
- more captions are l10n-compliant, the following can now be translated:
  - some static data
  - columns and fields
  - default database column values

- An example is provided: French, that improves what was done and provides a better coverage, including extensions

Bonus: fixes compatibility problems with `date-fns` version 2


okayawright:
I'm not sure anymore about the automatic column name translation; it can be really confusing: the physical field name in the database can be different from what the user see.
Maybe a better way to handle that would be not to rename field names on the fly but change the names within the database instead: e.g. let the user select the name of the field for "created_at" during table creation and then display the name as-is. But the problem remains for the core Directus tables; the overhaul is too complex in my opinion...

rijkvanzanten:
The fields collection in the database already has a translation column that's meant for this purpose. Instead of having hardcoded global translations for common names, it would allow the admin user to provide custom names in whatever language they want.

benhaynes:
Also, any automatic field creation (eg: `user_created`) can be disabled and created manually if you want to use your own names.

okayawright:
Ok, thanks. Let's drop this PR for now, it needs to be reworked then.

Issue #1375, opened 2019-03-06
M2O Interface


benhaynes:
No cancel button for modal (no way to exit)

The whole content area scrolls horizontally if too wide... but the search bar should not.

We should look into a better UX for selecting ONE. The checkboxes are a bit confusing.

srjrol:
> We should look into a better UX for selecting ONE. The checkboxes are a bit confusing.
I suggest an autocompleting list.  https://jqueryui.com/autocomplete/

benhaynes:
I like that — a header with filter/autocomplete would help a lot.

felskov:
@srjrol @benhaynes I've looked into it, and seems that allowing missing cancellation (#1452 ) is simply caused by an "action-required" prop on the v-modal tag.

I can create a pull request if you'd like, but it's a one-liner, and I haven't looked into the other things mentioned in this issue. So it might be faster for you guys to fix it on your own along with the rest of the things you're talking about here :-)

One thing I'd like to note: I'm using listing view preferences to apply a filter to list of selectable items, and sort by a specific column name. I'd be sad to see that feature disappear, if the entire UI was moved to an auto-completed input, but perhaps it'd be possible to maintain these configurations with such a solution as well :-)

rijkvanzanten:
> I've looked into it, and seems that allowing missing cancellation (#1452 ) is simply caused by an "action-required" prop on the v-modal tag.

This is true, but there is one downside to removing it: right now the interface immediately stages the value as soon as you select an item, so when you select an item and then hit 'cancel', the value is still staged 🤔 

> One thing I'd like to note: I'm using listing view preferences to apply a filter to list of selectable items, and sort by a specific column name. I'd be sad to see that feature disappear, if the entire UI was moved to an auto-completed input, but perhaps it'd be possible to maintain these configurations with such a solution as well :-)

Good point! I don't want to get rid of that functionality. That being said, it's not the most user friendly to setup right now.  I think I'd like to move to a setup where you can select what field to sort on, what fields to display (and how), and what filters to apply. Then we can use those three to render a dropdown autocomplete instead of using the modal with the nested table view.

rijkvanzanten:
This is fixed in https://github.com/directus/app/pull/1677

PR #1520, opened 2019-04-19
Add a cancel button when selecting a relational item

shealavington:
_Been going through and creating PR's today for issues that I know I've spotted — I seem to have found a related issue for most of them haha._

---

When you want to "add new" or "select existing",  you aren't "required" to do so, you should have the option to cancel.

From: https://github.com/directus/app/issues/1375
[Done-ish] No cancel button for modal (no way to exit)
_Hmm, just seen the comment about staging..._

[Done] The whole content area scrolls horizontally if too wide... but the search bar should not.

[Todo] We should look into a better UX for selecting ONE. The checkboxes are a bit confusing.

-

[Done] Padding is missing from some of the M2M / O2M / M2O. Looks weird

image


rijkvanzanten:
The problem here is with the expectation of "cancel": 

If you select an item, and then hit cancel, the item is still selected. I feel like this UX is not what people expect....

rijkvanzanten:
Instead, maybe we should rename `Save` to `Done` and make it always available?

shealavington:
That's potentially another good way of doing it.

skeemer:
> Instead, maybe we should rename `Save` to `Done` and make it always available?

That's more accurate to what is actually happening as well.

benhaynes:
Thanks Shea! So for the action button, it is always enabled and says "Done"?

shealavington:
Ehrrmm — This was something I was still looking at.

I've made some fixes, but the "Done" button being enabled and labelled correctly was not done. :thinking:

benhaynes:
Coolio!

@rijkvanzanten ?

shealavington:
I've just left the office.

I had a browse of the commit on my phone—it looks like the only changes from the merge were styling changes (which are all valid), for reference, it fixes a padding issue for the search-bar and makes it sticky , so side scrolling won't move it off the modal.

🙂

Given the time I have available, I'll revisit the original issue and see what I can do. 🙌🏻

rijkvanzanten:
This PR is just styling

PR #1521, opened 2019-04-19
Allow for optional time and deleting time entries

skeemer:
With the arbitrary adding of ":00" it doesn't allow for submitting empty time entries. It also needs set to null because the API assumes midnight if it is an empty string.


benhaynes:
Thank you @skeemer !!!

PR #1552, opened 2019-04-25
Fix notification icon name error on save

dev7ch:
image


rijkvanzanten:
Thanks 🙏🏻 this one has been bothering me for a bit 😋

benhaynes:
I had to look up the `!!` hahaha

Didn't know that was needed for `v-if`...

dev7ch:
`!!` it looks like that this avoids `undefined` errors on required and type defined `props`

rijkvanzanten:
Interesting. I was under the impression that props / v-if would consider undefined as falsy and therefore not use it. It does that correctly with null 🧐 oh well ¯\_(ツ)_/¯

Issue #1548, opened 2019-04-24
Can't add Interface on existing field


ThomasKliszowski:
# Bug Report

## Steps to Reproduce

1. Create a field directly on the DB (not in directus)
2. Click on this field in the Directus collection

## Expected Behavior

The "Update Field" form should be displayed, or the "Create Field" (as the directus_fields record does not exist) should not mark the name as already exists.

## Actual Behavior

You get a "Create field" form and you can't create it as the name already exists.

## Other Context & Screenshots

Field:
![image](https://user-images.githubusercontent.com/1174135/56681035-c0769500-66c8-11e9-8763-c788c1dded68.png)

Form:
![image](https://user-images.githubusercontent.com/1174135/56681060-cc625700-66c8-11e9-80cb-245804b9a921.png)

Error:
![image](https://user-images.githubusercontent.com/1174135/56681074-d4ba9200-66c8-11e9-9a64-bc4423f14edc.png)

## Technical Details

- Directus App: 7.3.0

rijkvanzanten:
@shealavington

shealavington:
Hopping on my computer now, just got in — please ensure you have the latest version, this was that bug that I found Rijk, I thought I'd already pushed a fix for it. 🤔

shealavington:
Oooo! Update, just re-read the issue, that's a nice edge-case, good spot. 
It's not what I was working on at-all. 

The screenshot shows an attempt to "Create" a field, which would correctly fail as it exists already, theoretically, you could just click the item and update the interface that way? @ThomasKliszowski

shealavington:
@rijkvanzanten What do we think about the following? 

* If the field-name isn't used: Can Continue
* If the field exists, without an interface: Can Continue: Warning
* If the field exists, and has an interface: Can't continue, must update existing.

As shown—You'd be warned that the field already exists, however, you can proceed.
On the other hand, does that not matter, should it just continue showing a '✔️'  ?
![image](https://user-images.githubusercontent.com/16869302/56686395-cb7ef480-66cb-11e9-96a0-729018a06981.png)

benhaynes:
So this flow is all for if the admin clicks "New Field" and tries to enter a field that already exists in the database (but doesn't have an interface yet)? I would say that we should show the blank field in the list (it looks like we area already?) and they should click on that to edit it.

I don't think there's a reason to use the New Field button for editing, so I think it should always say `"test" already exists` and _not_ let you go forward.

[...]

PR #1553, opened 2019-04-25
Updating database field not managed through directus

shealavington:
closes https://github.com/directus/app/issues/1548

**What does it do?**
Any database table with a column that doesn't have an interface selected in `directus_fields` should have a `manage` button shown in place of the interface, you must click this, and the system will select the closest matching interface based on the `datatype` of the column.

**Thoughts**
Though this now works well, I'm just wondering, **should have some kind of visual that shows it's processing / loading?** (This would also need adding to the Collections manage button too) as there's no instant feedback to show an action happening.

This takes longer than my last demo, as I now waiting for the request success, and then waiting for the store to update the collections before allowing the user to interact.

image


benhaynes:
Amazing!!

We have a button component that has an option for "loading spinner" (I believe, right @rijkvanzanten ?) — perhaps that could be used? 

1. "Manage" button clicked (row not clickable)
2. Turns into spinner button while processing (row not clickable)
3. Turns into interface name when done (row clickable)

This is great Shea — thank you for all the attention to detail!! 😄

shealavington:
🎉 Added loading icons — Now feels more responsive, rather than feeling like an action hasn't taken place.

https://www.loom.com/share/d98e1729b589491f863fd0e389f8343c

---
**Extra Bug Found — Maybe needs a new issue**
I've also found that you can navigate to: http://localhost:8080/settings/collections/unmanaged-collection and it will allow you to view the collection edit screen — Visiting an un-managed collection should refuse entry to the route until managed, or at-least fire the "manage" event on the collection before-entry.

src/components/form-inputs/button.vue
```diff
@@ -4,7 +4,7 @@
     :type="type"
     :disabled="disabled || loading"
     class="form-button style-btn"
-    @click="$emit('click')"
+    @click.prevent.stop="$emit('click')"
```
shealavington: @rijkvanzanten I've added "prevent.stop", as clicking the `Manage` button on the collections index was clicking through to the collection page. 

I don't think it will, but do you think or know if this will cause further issues with other places the "v-button" is used?

rijkvanzanten: I think it should be fine 🤷🏻‍♂️

rijkvanzanten: I think it should be fine 🤷🏻‍♂️

PR #1554, opened 2019-04-26
fix #1543 star interface

### LINKED ISSUE #1543: Stars interface. [bug][good first issue]
# Bug Report

## Steps to Reproduce

Click on an item with "X Stars"

## Expected Behavior

Preload the stars field with the amount of stars currently selected.

## Actual Behavior

Displays 0 stars selected by default.

Changing the stars does work, it just doesn't load the right default in. :sob: 

Selection_063
Selection_062

## Technical Details

- Directus App: 7.0.3

dev7ch:
closes #1543


src/interfaces/rating/input.vue
```diff
@@ -18,7 +18,9 @@
         :value="String(value) || '0'"
         @input="updateValue"
       ></v-input>
-      <span>/ {{ options.max_stars }}</span>
+      <span v-html="(!!value ? String(value) : '0') + '/' + options.max_stars">
```
rijkvanzanten: Does this have to be the unsafe raw html? Why not render this as escaped text in the span?

src/interfaces/rating/input.vue
```diff
@@ -1,13 +1,13 @@
...
   <div class="rating">
     <stars
-      v-if="options.display == 'star'"
+      v-if="options.display === 'star'"
       :options="options"
-      :rating.sync="rating"
+      :rating.sync="!!value ? value : rating"
       :readonly="readonly"
-      @update:rating="updateValue"
+      v-on:update:rating="updateValue"
```
rijkvanzanten: Isn't `@` the shorthand for `v-on`?

shealavington: I thought so which is why I used the @ there for the last fix I did. 🤔

dev7ch: upps yes

src/interfaces/rating/input.vue
dev7ch: ok done

PR #1562, opened 2019-04-27
Stop active palette item growing width.

### LINKED ISSUE #1523: Color palette width
Im not realy sure why this happens and also could't fix it but heres the problem:

The witdh of the button (column) doesn't match up with the others when selected even tho there is enough space and no need for having a larger width.

grafik

shealavington:
closes #1523 — Stops the width of the active column growing.

**There was another solution to this**: Using a negative margin, we could take the width of the item inserted (icon), negative margin half from the left, and half from the right, the only issue with this, is that if the size of the icon changes, you need to update the CSS too, therefore, I think `absolute` method was most efficient solution.

image
image
image
image


dev7ch:
```scss
i {
position: absolute;
left: 0;
right: 0;
bottom: 0;
top: 1px;
margin: auto;
}
```
Do you wanna center the icons in all circumstances? i assume logically it could be done a bit shorter and a bit more compatible. 

```scss
i {
position: absolute;
top: 50%;
left: 50%;
transform: translate(-50%, -50%);
}
```

benhaynes:
My only problem with translate is that with odd pixel heights it aligns it on sub-pixels and gets blurry. Otherwise it's a good way to reliably keep things centered. Perhaps the material icon set keeps them all at 24px squares (even)?

rijkvanzanten:
The icon will most likely not change from this check any time soon. I think we don't have to make this work for any icon

rijkvanzanten:
@shealavington lgtm. Feel free to merge this in once you've double checked @dev7ch's solution to check which one works better. @benhaynes has a good point about subpixel rendering potentially causing issues.

shealavington:
Good call @dev7ch, the following was the first thing I tried, however, I was getting the proceeding result (which doesn't work well). I'd rather use the pixel values than trying to guess the offset %. And as Ben mentioned, sub-pixel issues :) 

_(I went back to quickly make this change to screenshot lol)_

> ```scss
> i {
> position: absolute;
> top: 50%;
> left: 50%;
> transform: translate(-50%, -50%);
> }
> ```

image

rijkvanzanten:
The problem with percentages for these icons is also the fact that the bounding box of the icon is sometimes way bigger than the icon itself. The spacing for the material icons font can be ... all over the place really

dev7ch:
yes I also met some circumstances around that issue I think it fine as its is done;)

PR #1563, opened 2019-04-27
Fix positioning of notifcation route link.

### LINKED ISSUE #1544: Notifcations "Section" sits over my field names
# Bug Report

## Steps to Reproduce

Have a collection with lots of fields
go to edit the fields (index)
click the Information button to select which items to display
scroll down
notifications is `position: absolute` over my fields

## Expected Behavior

Not sit over my column names :P 

## Actual Behavior

I'm not sure on the intended behaviour here.

Selection_064

## Technical Details

- Directus App: 7.0.3

shealavington:
closes #1544 

This fixes the positioning of the notifcation route link, so that if you have lots of fields, it won't be absolute over the items.

* Button always fixed to the bottom of the screen (allows scrolling)
* Width is the same variable as the sidebar so the sizing is consistent between button and sidebar.
* Added `64px` padding to the element containing all the fields, 

This margin-bottom for the "fields div" was defined by the following calculation, 
* Top padding of the button (`20px`)
* Bottom padding of the button (`20px`)
* Biggest content  size (`24px`) 
* = `64px`

Unfortunately, CSS variables weren't being used to allow me to do a CSS calc for this — let me know if you think this should be re-factored to do the calculation using CSS calc, and I can create component variables for button padding and size to find size of the margin needed under the fields div?

**Before:**
image

**After:**
image
_Black line is my taskbar ;)_


rijkvanzanten:
Ideally, we rework the sidebar a bit so it's an overflowing div followed by this notifications button so we don't have to hack it on there with absolute or fixed. Anyway, this is a good stopgap until we have the time to 'properly' fix the sidebar markup 😄

benhaynes:
Awesome — that's how I would have fixed it... nice job and thank you Shea!!

Issue #1581, opened 2019-05-05
Select in Relation, does not show selected option.


henrikwirth:
# Bug Report

## Steps to Reproduce

1. Go to a Collection
2. Create or Edit a M2M/One2M/M2One Field 
3. Advance to to Relation Tab
4. See that the select fields don't show the selected option

## Other Context & Screenshots

![image](https://user-images.githubusercontent.com/10786260/57191456-45ae4500-6f26-11e9-9d16-1637d084f443.png)

## Technical Details

- Device: [eg: Desktop, iPhone6, etc]
- OS: OSX
- Browser: Chrome 73.0.3683.103
- Directus App: 7.3.1
- Directus API: 2.0.23
- Install Method: cloned `master` branch of `directus/directus`

https://github.com/directus/app/blob/6cd3d958def608b7a8f46ee0b6e2e6fb575f4a6b/src/components/form-inputs/simple-select.vue#L36-L66

This is happening because of the above code of simple-select.vue the `
el.children` is used, which only traverses 1 level down the do. The optgroups, which inherit the other options, have to be cosidered though. Therefore it would help to to `el.querySelectorAll("option")` to get all options down the DOM.

What I don't understand yet, is why it is needed to do getValueNames and save them all in an object to then get the value out of it again. Instead of just passing this.value in `valueText()`

@rijkvanzanten You did the last change there, maybe you can help me out here, so I can fix it.

rijkvanzanten:
> What I don't understand yet, is why it is needed to do getValueNames and save them all in an object to then get the value out of it again. Instead of just passing this.value in valueText()

`this.value` is the raw value that's saved in the DB, which can be different from what's shown in the dropdown 🙂

For example, in 

```html
<option value="rijk">Rijk van Zanten</option>
``` 

we want to show `Rijk van Zanten` in the valueText, not `rijk` (which would be `this.value`)

rijkvanzanten:
Fixed in https://github.com/directus/app/pull/1583

PR #1583, opened 2019-05-05
Fixes simple-select component to show selected value

henrikwirth:
fixes bug: #1581

Fixes the wrong traversing through the options. `element.children` only traverses one level down the DOM, which is not sufficient, because in some cases there are `optgroups`, which then have multiple options. Also it is important to add the `watch` to values again, otherwise the `valueNames` won't get updated when for example a new `junction_collection` is selected.


rijkvanzanten:
Thanks @henrikwirth!

benhaynes:
Thank you @henrikwirth — we really appreciate the help!!

PR #1584, opened 2019-05-05
Feature: Draggable check boxes

dev7ch:
this pr is an extension for the `checkbox` interface and allows you order the selected values by drag and drop

- new bool property `sortable` (default: `true`)
- the only allowed field type for the `checkbox`  interface is `Array`

**Preview Link**

https://deploy-preview-5--blissful-curie-6231b9.netlify.com/

(use default demo credentials)

**Questions**
- Default interface field type only and always to array ?
- Is this checkbox interface used for other field types on DB side?
- are there any edge cases which I have been out of scope (compatibility)?

Coz if i change the interface, from a dropdown to the checkbox i need to select the fieldtype to get it properly working. (but it closes directly after the interface is changed)


benhaynes:
Love this idea! I wonder if we should have an option to show it as a single column instead of wrapping... might make sense for longer labels or this sorting.

dev7ch:
@benhaynes yes single col mode is added 👍 

further I assume due there is `display:grid` in use its could be a good/needed to implement tooltips for longer labels and trim them by css to a fixed with.

What do you think @rijkvanzanten ?

https://deploy-preview-5--blissful-curie-6231b9.netlify.com/

src/interfaces/checkboxes/input.vue
```diff
@@ -1,53 +1,110 @@
...
-
       let selection;
-
-      // Conver the value to an array
-      if (typeof this.value === "string") {
+      // Convert the value to an array
+      if (typeof this.value === "string" && this.value) {
         if (this.value.includes(",")) {
-          selection = selection.split(",");
+          selection = this.value.split(",");
```
dev7ch: @rijkvanzanten why here is  `selection.split(",");`instead of `this.value.split(",");` ?

benhaynes:
Nice! Yeah, we should have these labels be overflow hidden with an ellipses (`...`), and each label should have a tooltip for the full text in case it doesn't fit. We have a tooltip attribute that should make this pretty easy.

src/interfaces/checkboxes/input.vue
rijkvanzanten: Very good question 🤔 It looks wrong doesn't it 🐛 

selection will always be `undefined` in that line, so I believe your change is better 👍

dev7ch: good thx :)

shealavington:
Maybe I'm being silly, I can't work out where this bool is to make it sortable?..

image

dev7ch:
Should be sortable on default, and only selected checkboxes are sortable

shealavington:
> Why would you like to disable the options?

Sorry — yes, you would be correct. Thinking about it again, the array would only be in the order they were checked anyway, so yes, don't worry about that one, always sortable wouldn't be an issue. :joy: 

> yes, would do it good, the cursor is set in a child component, that's why I didn't implemented it.

I think I follow what you're saying. Yeah, in that case, the only issue for me would be the UX and not realizing that they were sortable, or, is that my issue as a user? 

@benhaynes, you're good with the UX/UI stuff haha, thoughts on somehow showing that they're dragable?

dev7ch:
I think in general it would need a hint that's its draggable, I would give + 1 for cursor 😁

benhaynes:
I think it would be good to have an interface option for toggling `reorderable` since I could imagine a situation where the admin wants to keep values in a set order. To help show the difference and give affordance to the draggability, I would say that when the items are draggable we give them handles, like this:

Screen Shot 2019-05-09 at 10 42 58 AM

shealavington:
>  I could imagine a situation where the admin wants to keep values in a set order.

Ha, funnily enough, I thought exactly the same! Upon review however, it appears that by checking the box, the item is added to the array in the order that you checked the item anyway :man_shrugging: I don't see it making much difference as the admin has no impact..

dev7ch:
Yes i was thinking the same but:
- the order of the selectable options is defined in the interface options (ordable but norlt dragabble) 
- saved are only selected values (which are dragabble)

To make empty values dragabble o think your complete checkbox  have to be rethinked/rewritten.

So what we could do is to show an inactive drag handle on empty boxes (lighter gray) ok?

dev7ch:
But the ordering of keys in interface option is forced by your codeMirror settings

dev7ch:
Or i am missing sth? @benhaynes

benhaynes:
It makes sense that only the saved/selected values would be ordered when viewing the interface, but I think it makes sense to allow all to be draggable (the non-selected ones just won't save their order). So...

### Initial reorder/select
_Initial item order is based on interface option order. Users can reorder and select any items._
* [ ] Option One
* [x] Option Two
* [ ] Option Three
* [x] Option Four
* [x] Option Five
* [ ] Option Six

### Returning to saved values
_When viewing previously saved values, the list starts with the manually sorted values, then shows the other items based on interface options. Any items can now be selected/unselected/reordered._
* [x] Option Two
* [x] Option Four
* [x] Option Five
* [ ] Option One
* [ ] Option Three
* [ ] Option Six

@shealavington — I don't think we want to save values based on which you click first. We should probably update the logic so it's always based on the order of the items (either manual or initial).

Does all of this make sense to you guys? Hahah... complex to discuss over GitHub!

dev7ch:
Yes, makes all sense and that is 100% how it is implemented :)

I just disabled the ordering for unchecked, but imo it would be better to leave the unselected und fordable due its could lead into missunderstanding.

> It makes sense that only the saved/selected values 
yes that is the case.

> think it makes sense to allow all to be draggable
I think better would be: 1 Step: Select value. 2:drag selected value.

I wanted a add a very simple enhancement to make the upcoming editor toolbar sortable 🤣 🙈

shealavington:
> I don't think we want to save values based on which you click first. We should probably update the logic so it's always based on the order of the items (either manual or initial).

Agreed Ben, that would be the better approach, but does mean it would need a logic change to reflect the values in the original dataset rather than your click order, as it is currently.

I'm not sure it's something @dev7ch would like to/be able to work on? 😊

dev7ch:
> the non-selected ones just won't save their order

I think therefore is no change needed @shealavington, for the moment

lets make the enhancement of safe the order of unchecked in another PR :)
and here:

- update draggable icons
-  add more options to. drag more or less ;)

rijkvanzanten:
Just to echo some of the previous comments, lets:

- [ ] Make every checkbox draggable
- [ ] Add drag-handles
- [ ] Make the draggability based on settings instead of always-on

Also, I'm seeing a bit of a weird glitch where there's a wrong first item using the default options, but that might be unrelated to this PR:

image

dev7ch:
> Also, I'm seeing a bit of a weird glitch where there's a wrong first item using the default options ...

Yes that could be related to the thing, that the value in the DB of this field was not empty or not type of array ... was this after creation of a new checkbox field?

rijkvanzanten:
Yeah I just did a fresh pull of everything and added a checkbox interface with the default settings

dev7ch:
ok Done :)

I checked the empty issue mentioned about but it didn't occurs at mysids 🤔

src/interfaces/checkboxes/input.vue
```diff
@@ -36,6 +36,8 @@ export default {
     draggable
   },
 
+  props: ["disabled"],
```
rijkvanzanten: When / why / how / where is this used?

dev7ch: Probably to remove a undefined error, but I ll have to check tmr again

dev7ch: plz see this pr and comment for explanation.
https://github.com/directus/app/pull/1609

dev7ch:
hey hey @rijkvanzanten just in case u havent seen the resolved comments/changes, its already updated.

Happy Sunday 😃

rijkvanzanten:
I'm still seeing that weird issue where the first item is wrong:

image

This is on a fresh clone / fresh installation / fresh pull etc

These are the options saved to the DB. It doesn't show this first wrong item on the master branch, so there must be something going on in this change..

```json
{
    "choices": {
        "value1": "Option 1",
        "value2": "Option 2",
        "value3": "Option 3",
        "value4": "Option 4",
        "value5": "Option 5",
        "value6": "Option 6",
        "value7": "Option 7",
        "value8": "Option 8"
    },
    "wrap": false,
    "formatting": true,
    "draggable": true
}
```

Also, can we add a little drag-hand cursor when you hover the drag handle + add the dragging cursor when dragging?

dev7ch:
mmh strange, but why on the demo (https://deploy-preview-5--blissful-curie-6231b9.netlify.com) (which is a fresh build as well), if i just create a checkbox its not there 🤔 

It reminds me to some behaving, if the field schema type is not `Array` or theree arer soe values before in the database.

Could you maybe commit you local fresh clone to a detached repo (inluding package-lock.json) to i can clone it locally?
I just tested it now locally and on your Demo (in About collection) but it just  works fine to add a new checkbox field.

image

dev7ch:
Yes, **checkboxes label**, but the style looks like it got overridden in the checkbox component and i assume needs to be added there.

dev7ch:
@rijkvanzanten  could you please verify the weird thing happen if you paste this in your terminal?

```sh
git clone -b draggable-boxes git@github.com:dev7ch/app.git check && 
cd check && 
npm install && 
npm run dev

```

rijkvanzanten:
I'm cloning the PR right now using hub:

```
hub pr checkout 1584
```

I'll try doing a fresh clone from your fork

dev7ch:
Thank you!

rijkvanzanten:
Same thing 😨 

image

rijkvanzanten:
image

This is what it's returning in the console

dev7ch:
oh noo ... thanks for the log, i ll have a look and come back to you to Slack

src/components/form-inputs/checkbox.vue
```diff
@@ -103,4 +103,13 @@ label {
 input:checked + label {
   color: var(--darker-gray);
 }
+
+.sortable-box {
```
rijkvanzanten: Lets move this to the checkboxes interface instead of here

src/interfaces/checkboxes/input.vue
```diff
@@ -65,16 +124,100 @@ export default {
...
+    grid-template-columns: repeat(1, 1fr);
+  }
+  &.draggable {
+    max-width: 100%;
+    .sortable-box {
+      position: relative;
+      transition: opacity ease-in-out 0.2s, background-color ease-in-out 0.3s;
+      &.active {
+        &.ghost {
+          &.sortable-chosen {
```
rijkvanzanten: Do we need these levels of nesting? Makes it a bit hard to read

Maybe we can change it to

```
.active {}

.ghost {}

.sortable-chosen {}
```

Or if we need to explicit combination of those classes

```
.active {}
.active.ghost {}
.active.sortable-chosen {}
```

src/interfaces/checkboxes/input.vue
```diff
@@ -65,16 +124,100 @@ export default {
...
+              color: var(--dark-gray);
+            }
+
+            &:after {
+              color: var(--dark-gray);
+            }
+          }
+        }
+        margin-left: 12px;
+        :after {
```
rijkvanzanten: Adding `cursor` here would be ideal

src/interfaces/checkboxes/input.vue
```diff
@@ -65,16 +124,100 @@ export default {
...
+  }
+
+  &.single {
+    grid-template-columns: repeat(1, 1fr);
+  }
+  &.draggable {
+    max-width: 100%;
+    .sortable-box {
+      position: relative;
+      transition: opacity ease-in-out 0.2s, background-color ease-in-out 0.3s;
```
rijkvanzanten: Use variables for transitions

```suggestion
      transition: opacity var(--transition) var(--medium), background-color var(--transition) var(--slow);
```

(medium === 0.2s, slow === 0.3s)

src/interfaces/checkboxes/meta.json
```diff
@@ -59,7 +78,9 @@
       "formatting_comment": "Affiche les valeurs selon un format donné",
       "display_text": "Texte affiché",
       "value": "Valeur",
-      "option": "Option"
+      "option": "Option",
+      "single": "Single Column",
+      "single_comment": "Display checkboxes in single column"
```
rijkvanzanten: We should remove these if they're not in the correct language. The translations will fall back to en-US if the string is missing.

src/interfaces/checkboxes/input.vue
```diff
@@ -186,7 +186,7 @@ export default {
     max-width: 100%;
     .sortable-box {
       position: relative;
-      transition: opacity ease-in-out 0.2s, background-color ease-in-out 0.3s;
```
rijkvanzanten: Another thought, is the order of these correct? I thought it was property, duration, timing function. Eg `opacity 0.2s ease-in` vs `opacity ease-in 0.2s` 🤷‍♂

dev7ch: upps, yes u are right

PR #1589, opened 2019-05-06
Timeline layout

Nitwel:
ToDo:
- [x] Support any type of data in the container (where the user icon is at the moment)
- [x] Add loader when scrolling to the bottom
- [ ] Allow canging the content field without reloading the page
- [x] Make the User clickable (maybe needs to be done internal)
- [x] Fix the line-height of the User preview
- [x] Implement it to the activity preview

Preview:
Screenshot_2019-05-06 Timeline


src/layouts/timeline/meta.json
```diff
@@ -0,0 +1,110 @@
+{
+  "name": "$t:timeline",
+  "icon": "date_range",
+  "translation": {
+    "en-US": {
```
rijkvanzanten: Seeing that this is a core layout, feel free to move this into the "global" language file (`/src/lang/en-US.js`). I'm planning on doing the same thing for all system interfaces as well.

Nitwel: Okay, should I do it also for the calendar interface?
And when yes, in this merge?

Also tought of it because the calendar and timeline interface will use mostly same translations.

rijkvanzanten: Feel free to do it for the calendar too, yeah! 

> Also tought of it because the calendar and timeline interface will use mostly same translations.

That's one of the reasons why I want to move them into the /lang folder 🙂

Nitwel:
directus/api#964

PR #1591, opened 2019-05-06
Update Bug_report.md

rijkvanzanten:
People have been filling out the bug reports very poorly. A half / badly filled out template is harder to work with than no template at all. I've also noticed that the people with the best formatted reports don't use the template structure either.. Maybe we can try it out with a less "formatted" format?


benhaynes:
I'm fine with this, but feel like we can add a note about the most important info that helps us:

```
Some useful details to include: your version of Directus, stack/server info, screenshots, logs, or even database dumps.
```

Just having that as a reminder (instead of a structured "todo" list) might help people draft better reports.

henrikwirth:
I actually found some guiding headings useful :)

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -6,40 +6,21 @@ about: Create a report to help us improve platform stability
 ---
 
 <!--
-For security issues, please email support@directus.io directly.
```
shealavington: I think this could be a good change, if it doesn't work then we can always revert—as mentioned, when filled out poorly, it leaves a 'junk' template to look through, which is annoying when trying to see the value of the issue.

However, I would keep the line about security issues being emailed.

shealavington: I think this could be a good change, if it doesn't work then we can always revert—as mentioned, when filled out poorly, it leaves a 'junk' template to look through, which is annoying when trying to see the value of the issue.

However, I would keep the line about security issues being emailed.

shealavington:
Looks good to me, originally I thought the line about security issues was missing, but I see it there :wink: haha!

rijkvanzanten:
@henrikwirth It depends per person. Lets merge this in as a tryout. If we see that the issues improve over time, we can commit to something like this and use it across the organization, otherwise we'll revert 🙂

Issue #1594, opened 2019-05-06
Interface: Divider - Required is set when field gets edited


henrikwirth:
# Bug Report

## Steps to Reproduce

1. Add the Divider Field to a Collection
2. Edit the Divider Field and Save
3. Try to create an item of this Collection
4. See error: `divider: This value should not be blank.` 

Clearly the required flag is saved to 1 (true) after editing the divider. Could this be a problem with all interfaces, that do not show the required field?

## Other Context & Screenshots

![image](https://user-images.githubusercontent.com/10786260/57255163-632cfd00-7053-11e9-8d6a-c57fe2049715.png)

## Technical Details

- Device: Desktop
- OS: MacOS 10.14
- Browser: Chrome 73.0.3683.103
- Directus App: next.directus.app

fnicastri:
The divider component has the required prop set to true.
Can't find where this prop come from, any hints?

rijkvanzanten:
When creating a new field, the default value for required is `true`. I'm assuming that the required state of this interface in this instance comes from that default value for `required`

benhaynes:
Could we override this with the `recommended` object in `meta.json`?

rijkvanzanten:
That's exactly how @henrikwirth fixed it in https://github.com/directus/app/pull/1595 😄

rijkvanzanten:
Closed in https://github.com/directus/app/pull/1595

PR #1595, opened 2019-05-06
Fix for Divider Interface

henrikwirth:
#1594

Adds the required field to the recommended options and therefore makes it possible for extensions, to predefine the required state. Also changes the required toggle to be only disabled in the case of an alias use. This gives more transparency while still stopping the user from acting. Because the default state of the required field is true, the value seems to be updated to true when the divider field gets edited. This should be fixed now. 

Yet, it is unclear to me, why this effect does not appear when the divider is created. I have the feeling, that the value should be properly pulled, when editing the divider field. Then my solution would be rather a workaround than an actual fix. Still, the recommended->required option could be a nice thing.


rijkvanzanten:
Great addition @henrikwirth, thanks. As you've probably noticed while working on this, the field-setup component got quite big and complicated over time. That will be one of the first things I'll refactor after next week. I will definitely keep the recommended->required option though 😄

henrikwirth:
Yes it was super hard to find my way through it :D 
Also i highly suggest to show a disabled toggle for required instead of just not showing him. I feel this gives more transparency for the user. He then knows the value is there and it is set, but for that interface it has to be false.

Issue #1580, opened 2019-05-04
Many-To-One relation field not disabled if readonly in role


fnicastri:
# Bug Report

## Steps to Reproduce

Define a collection with a Many-To-One relation.
Mark the collection readonly for a specific role
Login with a user with that role

## Expected Behavior
The relation dropdown should be disabled
## Actual Behavior
The relation dropdown is still active even if you can't save the changes
## Other Context & Screenshots
![image](https://user-images.githubusercontent.com/5978520/57178276-0a503f80-6e6e-11e9-926a-129649edd384.png)

## Technical Details

- Directus App: latest 6cd3d958def608b7a8f46ee0b6e2e6fb575f4a6b

fnicastri:
Think I've found the issue, I'll send a PR in the next few days (out of town for a couple of days)

PR #1596, opened 2019-05-07
Fixes readonly on many-to-one field #1580

fnicastri:
Add the disabled property to the v-select in many-to-one interface to respect readonly value #1580


benhaynes:
Thanks @fnicastri !! Busy week, but we'll take a look ASAP!

fnicastri:
👍

fnicastri:
thanks guys

Issue #1605, opened 2019-05-09
Adding storybook.js for isolated component testing


henrikwirth:
# Feature Request

### What is storybook?
> Storybook is a user interface development environment and playground for UI components. The tool enables developers to create components independently and showcase components interactively in an isolated development environment.

![image](https://user-images.githubusercontent.com/10786260/57470378-f9963400-7288-11e9-97a0-2a3764a0dd91.png)

## What problem does this feature solve?
It makes it possible to test components, interfaces, layouts etc. isolated from other content. Also it can serve as a showcase of what components there are and how they work. Also, how they might look like in different conditions (different background, with or without main.scss ...) I think there is many ways, how you could use it. It would be up to our creativity.

There is also plenty of addons, that tackle different problems. Check it our here: https://storybook.js.org/addons/

## How do you think this should be implemented?
I already setup a working implementation on a branch which I will link through a pull-request. The main task will be to decide, what components and layouts should be showcased and under what conditions should they be implemented.

It would be great to somehow have a pipeline, that builds the static version and serves it somewhere. Could be also interesting to introduce some versioning to it and therefore see the development of things.

I would love to leave the rest for discussion :)

## Would you be willing to work on this?
Indeed

benhaynes:
Interesting! Any idea how well our components would work within a platform like this @rijkvanzanten?

henrikwirth:
@benhaynes feel free to check out my example. Not perfect, but i guess it shows, that it can be used ...
just `npm i` and `npm run storybook`

https://github.com/henrikwirth/app/tree/feature/introduce-storybook

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

PR #1606, opened 2019-05-09
[WIP] Introducing storybook to directus

henrikwirth:
Feature Request: #1605 

This should serve as the initial setup for storybook. Important will be a proper folder structure and a way of integration, that doesn't effect the rest of the build. Some example should show, how Storybook can be used and they should serve as template for all coming stories. Not all components have to be integrated directly. This can be done step by step, while working of refactoring.

### Initial Setup
- Folder Structure
```
•
└── src
    └── components
        └── button
            ├── button.vue
            ├── button.stories.js
            └── button.readme.md
```
- Load all vue components from `globals.js`, so they don't have to be imported on every story
- Load main.scss as base of all the styling
- The note-plugin should be implemented, so that the `component.readme.md`s can be imported

### Addons
Can be found here: https://github.com/storybooks/storybook/tree/master/addons/ or here: https://storybook.js.org/docs/addons/addon-gallery/

I would suggest the following:
- Note or Readme (Markup based Readme/Note that can serve as documentation)
- a11y (For easy accessibility checks)
- Knobs (lets you change properties live)
- Backgrounds (Lets test every component on different backgrounds)
- Storysource (shows how story is implemented)
- Actions (lets you test events)
- Viewport (to test on different screensizes)

There is many more addons, that could be useful, depending on how we want to utilize storybook

### Update Docs

- node-gyp might need some extra setup, which should be mentioned in the docu (https://github.com/nodejs/node-gyp#installation)
- node-sass can make problems: `npm rebuild node-sass` can help

## Ideas/Suggestions

- The Interfaces at the moment have a preview Page. Maybe it would be possible to use this functionality and get it all into Storybook, so that it is not redundant
- The storybook build could be build and served as a static page. Could be easy with something like netlify.


henrikwirth:
Folder structures:
image

The last one, separates it more from everything production related, where-else the first option would kind of force everyone to write a story for each component. A component could then be seen as one package with its component code, representational story code and maybe tests.

rijkvanzanten:
Lets go with the `stories inside component directory` option. As we've discussed on Slack before, we'll end up with

```
.
  src
    components
      button
        button.readme.md
        button.stories.js
        button.vue
```

henrikwirth:
Somehow we managed to have problems with this branch and therefore setup a new one #1644

PR #1609, opened 2019-05-10
fix missing disabled props in many2one input

dev7ch:


rijkvanzanten:
All the interfaces get their props from [the same mixin](https://github.com/directus/extension-toolkit/blob/master/mixins/interface.js). Why should only the many to one have the additional `disabled` prop?

dev7ch:
mmh yes I was wondering as well about that but I got an undefined error there, get back tmrw about this

dev7ch:
But here is `disabled` in the code which i didnt see in the mixin, was there a BR?

https://github.com/directus/app/blob/ec6219d7ffbcf57a974502843c02bf99069c9948/src/interfaces/many-to-one/input.vue#L16

dev7ch:
so then i assume the `disabled` is either missing in mixins or wrong in m20 interface @rijkvanzanten

rijkvanzanten:
We used to have both disabled and readonly (which both did a slightly different thing), but those two got merged together into `readonly`. This check for `disabled` was never removed when that change happened.

dev7ch:
Mmh then i would suggest add propbas in the PR to avoid errors for now and remove it when interface or/and module will get a closer look.

rijkvanzanten:
I don't agree, the `disabled` prop is never passed to the interface's `input` component so it will always be `undefined`.

https://github.com/directus/app/blob/ce36fb79bbf3965ed06228c8a0be4f768c1a868d/src/components/extension/interface/input/input.vue#L2-L21

Instead, we should remove the check for `disabled` variable, eg

```
// before
:disabled="disabled || readonly" 

// after
:disabled="readonly" 
```

dev7ch:
ok yes ofc that's the better way, but I was unsure if there are now functions (hidden) which depend on it

dev7ch:
its removed here @rijkvanzanten

PR #1614, opened 2019-05-13
tag-component

benhaynes:


rijkvanzanten:
Is a tag always a button, or will we re-use this component for other things as well?

src/interfaces/tags/input.vue
```diff
@@ -141,16 +141,10 @@ export default {
...
   button {
-    transition: var(--fast) var(--transition);
-    margin: 2px;
-    padding: 2px 4px 3px;
-    background-color: var(--gray);
-    color: var(--white);
-    border-radius: var(--border-radius);
-
     &:hover,
     .user-is-tabbing &:focus {
```
rijkvanzanten: We have the hover specific styles here, but the other styles in the tag component. Can we move the hover styles there too?

benhaynes: I was thinking of the tag component as a "base", but there are so many ways you might want to interact with it... and creating props/flavors of each might make it overly-complex. That's why I went with this "override" direction.

You might want a specific background color on hover (and therefore a specific foreground color), you also might want to change the cursor, and I wasn't sure about how the `.user-is-tabbing` would be handled if it were moved into the base component.

benhaynes:
I suppose it could be used for anything that is visually a tag (not always clickable). Would we want a conditional to make it a `button`/`span` in that case?

rijkvanzanten:
Yes

PR #1615, opened 2019-05-13
Make Lodash globally available under _ 

rijkvanzanten:
It was already "globally" available by using `this.$lodash` inside of Vue components. This PR replaces that `this.$lodash` alias by the more common variable name `_`. This should make it easier to understand where certain methods are coming from, and it allows custom extensions to use Lodash in JavaScript files without having to import it separately.


benhaynes:
Will previous usage of `this.$lodash` still work?

rijkvanzanten:
In this case I replaced all the instances of `this.$lodash` with `_` and removed the old alias to prevent confusion between the two.

benhaynes:
Makes sense. Is there a chance that custom extensions are using the old syntax and will break? In other words, is this a breaking change?

rijkvanzanten:
Docs update: https://github.com/directus/docs/pull/145

Issue #799, opened 2018-09-08
Data Migration Tool


benhaynes:
#### :rotating_light: MIGRATED FROM REQUESTS.GETDIRECTUS.COM :rotating_light: 

# Data Migration Tool
### :+1: = 1
*Created 4 weeks ago by @dynamikapps*
 

-----
### @dynamikapps – :+1: (4 weeks ago)
If the person managing the client app does not have the technical background to migrate a DB and would like to move data from a table/collection/item from let say a staging envr to a prod envr then it would help to have that feature. Or even if they would like to move just the schema structure (ei. person table - name, address, age, postal etc...) from one envr to another would be nice as well.

Including general Import/Export and different formats (JSON, SQL, XML, CSV, etc)

Thanks

rijkvanzanten:
Duplicate #711 #765

PR #1633, opened 2019-05-16
Feat - Repeater

cesasol:
This is the continuation from [#799](https://github.com/directus/api/pull/799). I think is mostly ready.

### DIFF
--- a/src/interfaces/repeater/group.vue
+++ b/src/interfaces/repeater/group.vue
@@ -8,7 +8,9 @@
       <div @click="collapsed = !collapsed">
         <v-icon :name="collapsedIcon" :class="collapsedIcon" size="18"></v-icon>
       </div>
-      <div><v-icon name="delete" class="delete" size="18"></v-icon></div>
+      <div>
+        <v-icon name="delete" class="delete" size="18"></v-icon>
+      </div>
     </div>
     <div v-show="collapsed" class="repeater-group__body">
       <div v-if="schema.type === 'object'">
@@ -21,17 +23,13 @@
           />
           <repeater-group
             v-else-if="typeof value[key] !== 'undefined'"
-            style="border-color: red;"
             v-model="value[key]"
             :schema="field"
             :key="key"
           />
-          <span :key="key + 'span'">^value.{{ key }}</span>
         </template>
-        <hr />
       </div>
       <div v-else-if="schema.type === 'array'">
-        {{ value }}
         <template v-for="(group, i) in arrayFields">
           <repeater-field
             v-model="value[i].value"
--- a/src/interfaces/repeater/group.vue
+++ b/src/interfaces/repeater/group.vue
@@ -126,24 +126,24 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped>
+<style scoped>
 .repeater-group {
   display: block;
   border: 1px solid var(--blue-grey);
   margin-bottom: 1rem;
+}
 
-  &__header {
-    padding: 0.5rem;
-    justify-content: space-between;
-    display: flex;
-    border-bottom: 1px solid var(--blue-grey);
-    label {
-      padding: 0 1rem;
-      flex: 1 0 auto;
-    }
-  }
-  &__body {
-    padding: 1rem;
-  }
+.repeater-group__header {
+  padding: 0.5rem;
+  justify-content: space-between;
+  display: flex;
+  border-bottom: 1px solid var(--blue-grey);
+}
+.repeater-group label {
+  padding: 0 1rem;
+  flex: 1 0 auto;
+}
+.repeater-group__body {
+  padding: 1rem;
 }
 </style>
--- a/src/interfaces/repeater/group.vue
+++ b/src/interfaces/repeater/group.vue
@@ -1,0 +1,151 @@
+<template>
+  <div class="repeater-group">
+    <div class="repeater-group__header">
+      <div>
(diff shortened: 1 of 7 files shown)


cesasol:
I'm not sure if the reorder functionality should wait for other reorder like efforts or use my own way with https://sortablejs.github.io/Vue.Draggable/#/simple

benhaynes:
Thanks @cesasol — let's see what @rijkvanzanten thinks on Monday!

src/interfaces/repeater/meta.json
```diff
@@ -0,0 +1,107 @@
...
+        "template": [
+        ]
+      },
+      "default": [
+        {
+          "type": "object",
+          "$key": "columns",
+          "title": "Columnas",
+          "properties": {
+            "titulo": {
```
rijkvanzanten: Can we make this English?

src/interfaces/repeater/meta.json
```diff
@@ -0,0 +1,107 @@
...
+              "title": "Subtitulo de columna",
+              "$key": "subtitle",
+              "type": "field",
+              "interface": "text-input",
+              "interfaceOptions": {}
+            },
+            "tarjetas": {
+              "type": "array",
+              "$key": "cards",
+              "title": "Tarjetas",
```
rijkvanzanten: Lets use `$t:tarjetas` instead of a hardcoded single language

src/interfaces/repeater/meta.json
```diff
@@ -0,0 +1,107 @@
...
+            "titulo": {
+              "title": "Titulo de columna",
+              "type": "field",
+              "$key": "title",
+              "interface": "text-input",
+              "interfaceOptions": {}
+            },
+            "imagen": {
+              "title": "Imagen de tarjeta",
+              "$key": "image",
```
rijkvanzanten: What does this `$key` do here?

src/interfaces/repeater/input.vue
```diff
@@ -0,0 +1,126 @@
...
+    return {
+      currentItems: [],
+      isValid: true,
+      originalValue: ""
+    };
+  },
+  computed: {
+    /** @returns {ReplicatorConfig[]} */
+    schema() {
+      const s = Array.isArray(this.options.schema) ? this.options.schema : [this.options.schema];
```
rijkvanzanten:
```suggestion
      return Array.isArray(this.options.schema) ? this.options.schema : [this.options.schema];
```

src/interfaces/repeater/input.vue
```diff
@@ -0,0 +1,126 @@
...
+      return s;
```
rijkvanzanten:
```suggestion

```

src/interfaces/repeater/group.vue
```diff
@@ -0,0 +1,165 @@
...
+      <label>{{ schema.title }}</label>
+      <button @click="collapsed = !collapsed">
+        <v-icon :name="collapsedIcon" :class="collapsedIcon" size="18"></v-icon>
+      </button>
+      <button v-if="!confirmDelete" @click="confirmDelete = true" key="delete0">
+        <v-icon name="delete" class="delete" size="18"></v-icon>
+      </button>
+      <button v-else class="confirm-delete" @click="$emit('delete')" key="delete1">
+        <v-icon name="delete" class="delete" size="18"></v-icon>
+        ¿Are you shure?
```
rijkvanzanten: Lets replace this with `$t()` and add a translation for this in `en-US.js`.

rijkvanzanten:
This is already working pretty nice! The setup is a bit hard though... This will need a solid documentation @benhaynes

src/interfaces/repeater/meta.json
cesasol: It's to track the type of field from the DB, every field but the direct childs from an object set are saved as `{"value": "", "$key": ""}` so it may be recognized later for the proper field

src/interfaces/repeater/meta.json
cesasol: It is just an example so I don't think that this value should be translated, however I can set it up on english.

cesasol:
@rijkvanzanten  I do agree the setup is a little complex I can write the documentation this week but as my english is not great maybe someone can help me to correct the grammar and redaction.

benhaynes:
This is looking great!!

We definitely need to keep all the example (dummy placeholder data) in English for now, with keys translated as Rijk mentioned. Let me know if you need help with this...

As for setup, we have a few options:
1. We re-use the Field Setup flow to create these fields. We'd need to make sure that component works with this... and would need to figure out how to handle 2+ stacked overlays at the same time.
2. We could leave the _complex_ fields for the **relational repeater** interface. Since those fields will be setup in Settings (like normal). Then we could find a more simple way to setup these fields.

I'm down to discuss! We'll definitely need _some_ sort of interface for setting up that `schema` option. I'd love to hear other recommendations!

rijkvanzanten:
This interface is virtually identical to the one @Nitwel made. However, this provides one piece of functionality that the other one doesn't: choosing what "type" of row to add. 

EG, the current repeater only allows you to create multiple instances of the same row. This repeater aims to add in support that the user can _choose_ what "type" of row to add. Seeing that the rest of the interface is virtually identical, I think we should combine efforts and try to implement that row-choosing-function as an option of the existing repeater field.

@Nitwel thoughts on how viable this is? I think we should be able to merge @cesasol's great work on top of the "1-dimensional" repeater.

I'll close this PR as we're moving to making this functionality an option of the existing repeater.

PR #1639, opened 2019-05-20
Update input.vue

dev7ch:
I found a small bug here, if there are no empty "ghost" values, this avoid throwing undefined error and ensurer there is at least one value


benhaynes:
Thanks @dev7ch!!! 😄

Issue #1575, opened 2019-04-30
Nav Override Interface


benhaynes:
This is a mock design for a dedicated system interface to control the Navigation Override. @rijkvanzanten you have a `collections` include, but we should also allow for individual collections to be added (as seen here) so that they can be reordered (non-alphabetical) or differently colored.

You are also using groups, while this uses dividers — we'll want to look into the best way to accomplish this. I like groups because they allow us to collapse (in the future)... but dividers are easier to setup since it keeps things linear. 

![Directus 8 – Settings Role](https://user-images.githubusercontent.com/522079/56985604-b506ec00-6b56-11e9-9031-f42c5c2b1e54.png)

Ref: https://github.com/directus/app/issues/125

```
[
    {
    	"title": "Collections",
        "include": "collections"
    },
    {
    	"title": "Style Guide",
        "links": [
            {
            	"name": "Directus",
                "path": "https://example.com/style-guide",
                "icon": "link",
                "target": "__blank",
                "color": "green"
            }
        ]
    },
    {
    	"title": "Favorites",
        "include": "bookmarks"
    }
]
```

rijkvanzanten:
A group can still work in this linear setup though. A group is nothing more than the selection of rows between two dividers

rijkvanzanten:
Also, links can go to internal pages as well:

```
           {
            	"name": "Directus",
                "path": "/collections/movies",
                "icon": "link",
                "color": "green"
            }
```

benhaynes:
That's exactly what I was thinking — it's a system interface so we can decide how it processes and saves the JSON. 

One thing that some people might find annoying is that the icons will be saved on the collection detail page, not the nav override. I think it might be a little confusing to have both.

Also, would want to add away in the JSON to allow for coloring the includes and changing their icons.

ceruulean:
Maybe for now the icon can just get the name of the one already set in collections?

ceruulean:
You want the below setting to make everything green?
`           {
            	"title": "Section Title",
                "includes": "collections",
                "color": "green"
            }`

Also would "target" be for an "Open in new window?" option? (i don't think iframes are very common but still has use cases like opening the youtube source page)

benhaynes:
Yes, that would turn the whole collections area green. If you wanted individual collections to be different colors you would need to add them individually.

Yes, the Target key would be for opening in a new tab or the same page.

[...]

PR #1642, opened 2019-05-20
Repeater Interface

### LINKED ISSUE #1570: Add "status mapping" interface
@theharshin posted a great mock for this (ClickUp) that would make visualizing and setting up the status mapping much easier.

Nitwel:
ToDo
- [x] make boxes draggable
- [x] add documentation
- [x] add the display view
- [x] polish up design
- [x] add a limit/amount option
  - [x] dont show the add button when limit = 1
- [x] add it to all / some Interfaces with json-config:
  - [x] Status
  - [x] Dropdown*
  - [x] Multiselect*
  - [x] Button Group*
  - [x] Radio Buttons*
  - [x] Status Mapping
- [ ] Make this a core component
  - [ ] could be used for the relational repeater and for the navigation override.
  
*Need further dicussions because the format is simpler like for the dropdown but could also be implemented as Json format ot add the feature that the interface understands these structures too.

closes #1570


Nitwel:
Only somehow there is something not working for collection display when using the new interface to setup the status interface. The single item view (when clicking on the +) is working fine tho.
grafik

rijkvanzanten:
It looks like there's a type inconsistency. Can you compare how this interface saves the value to the database compared to how it's saved by default? I'm guessing one is a string and the other a JSON object.

Nitwel:
The Json interface saves the string as an JSON object:
grafik

The Status-Mapping Interface does the same:
grafik

Thats what confused me because it should be the same.

rijkvanzanten:
Hmm strange. Maybe the app does another JSON.parse on the value that's already an object?

Nitwel:
Ah, I think I got it. I must have mest up the config of the status interface. The field type shouldn't be JSON
Screenshot_2019-05-21 Settings Editing Item Status

Nitwel:
Nice, nested Status mapping also works. 🎉
This will be very usefull for button groups or to configure the toolbar of the WYSIWYG editor.
Screenshot_2019-05-22 Editing Item Status

rijkvanzanten:
This is awesome! Good work @Nitwel. You're turning the status mapping interface into a full fledged visual JSON editor 😄

Nitwel:
Ah, yep, about that. What about a better name than `status mapping`?
I dont know any better but I think the name does not perfectly describe the functionality of the interface.

rijkvanzanten:
@benhaynes will most definitely have an idea about that haha

In terms of functionality, it's most like a _repeater_ I think. You can setup an initial row structure after which the user can repeat this row as many times as they like.

Nitwel:
Some kind of JSON based repeater and not a repeater that stores the data raw in a new table of the database. What about `JSON - Repeater` or something simular?

Nitwel:
Theres one thing that is that annoying me. The Button group interface for example uses arrays for its nesting and not objects (same with the text-editor), it would be super nice if it would support both ways so I dont have to make the "JSON - Repeater" even more complex. I would add this feature to the button group if you also think its a better way of doing it.

rijkvanzanten:
Can you provide an example of both the structures you're talking about?

benhaynes:
I think we would go the other way... this is a `repeater`, and the other would be `repeater-relational`. That would match how we did: https://github.com/directus/app/tree/master/src/interfaces/checkboxes-relational

Nitwel:
The issue I had is that the button group interface uses arrays too like here:
grafik
The only small disadvantage that now exists is that buttons wont be able to be "groupless" anymore like here: 
grafik
But would need a container like one on the first image.

Nitwel:
The interface is now complete by my side. 😄 
Theres still one thing to discuss. When configuring the repeater interface, you have to assign a interface type for each field. At the moment I just added some basic fields but it would be nice to have this list automaticly generated. To do this, we would need an interfaces interface which lists all interfaces as a Dropdown or such to select an Interface.
What are your thoughts about this  @rijkvanzanten ?

rijkvanzanten:
An interfaces interface sounds like a relatively simple one to make 🙂 We can read all the interfaces from the store. It will be a giant list though..

Nitwel:
Or what about having the same select like for normal fields but just as an interface?
Because it would be annoying to do this manual.

benhaynes:
We were talking about adding the category to each interface's `meta.json` — if we have that we can group them in a system dropdown. I don't think the list will be too crazy. Also, a system dropdown will give the best experience on touch.

Nitwel:
I have added an `Interfaces` interface and its not "too" large so I think its the best way of doing it at the moment.
grafik

rijkvanzanten:
@Nitwel Can you open a separate PR for the interfaces interface? So we can improve / tweak it without it being cluttered by the changes in this one.

Nitwel:
Ugh, yea wait a sec.

Nitwel:
Here: #1668

rijkvanzanten:
We can optimize the interface once the interfaces have their groups listed in meta.json @benhaynes. This is workable for now 👍

matoware:
Sorry for that dumb question but: How do i get informed when i should update directus via **git pull** to have that repeater-functionality included?

Nitwel:
You can check out twitter for latest updates. https://twitter.com/directus/
Rijk planned to merge this into the master today so it will probably come with the next release.

matoware:
Are there also notifications about that in the Slack channel?

Nitwel:
Yep, in the channel "notifications" but it could be a bit overwhelming somethimes because it shows all kinds of things.

Nitwel:
Ah, wait. This interface still has no proper display!
Any suggestions / ideas @benhaynes ?

benhaynes:
I think we should do the same as other interfaces (eg: relational) that have multiple items: show the item count.

`3 Items`

Looking at the others (`files`, `m2m`, `o2m`) they all have some crazy code within `display.js` which I think shows an overlay of content when you hover. That could be added later.

rijkvanzanten:
@matoware If you hit Watch in the top right of the repository page, there is an option that will notify you when there are new releases 🙂

Nitwel:
When hovering over it, it shows a list of the items too.
grafik

The next thing I had in mind for today is to make the repeater interface into an internal one but to still have it as an interface. (Like the text-input interface ist mostly just a v-input)
This allows us to use this feature in a nav override for example. 
See here: #1575

src/interfaces/repeater/display.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <v-popover trigger="hover" v-if="valueFields && valueFields.length > 0">
```
rijkvanzanten: Could this use this new component? https://github.com/directus/app/pull/1672#event-2365523082

rijkvanzanten:
> This allows us to use this feature in a nav override for example.

That page itself is a v-form that uses directus_fields :) Having this interface already enables us to use it for the nav override by updating the database in a migration.

Nitwel:
Well, if needed this could be added later on but for now everything should work properly.

benhaynes:
Hey @Nitwel @rijkvanzanten — I've made some style/default updates here. Some other thoughts on this interface:

* Should we have a confirmation prompt for deleting items? You _could_ have a lot of data in there, and that button is right next to the "expand" button.
* How do you update interface options? For example, if you wanted to use a 'dropdown" interface, there is no way to setup the options

Nitwel:
We should definetly include an options field for release so just in case an interface needs this, it is right there.

benhaynes:
How would that be entered though? Would all interface options be shown or would it be a JSON field where they need to enter manually?

rijkvanzanten:
We might be able to render a nested `v-form` based on the meta.json of the interface you selected 🤔

Nitwel:
I think the best way to do this is to temporaly add just a plain JSON input for the options and later redo the setup process to use the v-field-setup to make this all work atomaticly but first we have to get the v-field-setup redone a bit. ^^

rijkvanzanten:
Hmm, I'm not too big of a fan of inline JSON fields. They tend to be too hard for regular people to setup.

rijkvanzanten:
We could invent an interface-options interface that renders the options for a given interface based on the value. It would be a system interface, but at least that prevents the inline JSON 🙂

Nitwel:
Weeeelll, have you seen that someone who has 0 clue of databases & directus decided to just create some fields and do nothing more with it. XD
Altho just an empty JSON spot is 100% ugly and I would hate so setup this my own but at the moment theres not really a better solution for it. And because the Repeater interface is just a nested form, we would have to add code onto the v-form or introduce something new that fields can react to other fields.

benhaynes:
I like the idea of an `Interface Options` system interface. Seems like it would be used in a lot of our more _meta_ interfaces (like this one).

Nitwel:
Well yes, it would be used for all interfaces that need some sort of fields configuration. 😄

justrealmilk:
How difficult would it be to alter the behaviour of the "add item" button in post mode to allow a user to select **one** of the repeater interfaces as opposed to getting a group of all of them?

rijkvanzanten:
@justrealmilk I feel like that would be a different interface. I don't think the current UI of the "boxes" is the best suited if you have multiple "rows" of individual different fields.

PR #1644, opened 2019-05-20
Storybook

henrikwirth:
Feature Request: #1605 

This should serve as the initial setup for storybook. Important will be a proper folder structure and a way of integration, that doesn't effect the rest of the build. Some example should show, how Storybook can be used and they should serve as template for all coming stories. Not all components have to be integrated directly. This can be done step by step, while working of refactoring.

### Initial Setup
- Folder Structure
```
•
└── src
    └── components
        └── button
            ├── button.vue
            ├── button.stories.js
            └── button.readme.md
```
- Load all vue components from `globals.js`, so they don't have to be imported on every story
- Load main.scss as base of all the styling
- The note-plugin should be implemented, so that the `component.readme.md`s can be imported

### Addons
Can be found here: https://github.com/storybooks/storybook/tree/master/addons/ or here: https://storybook.js.org/docs/addons/addon-gallery/

I would suggest the following:
- Note or Readme (Markup based Readme/Note that can serve as documentation)
- a11y (For easy accessibility checks)
- Knobs (lets you change properties live)
- Backgrounds (Lets test every component on different backgrounds)
- Storysource (shows how story is implemented)
- Actions (lets you test events)
- Viewport (to test on different screensizes)

There is many more addons, that could be useful, depending on how we want to utilize storybook

### Update Docs

- node-gyp might need some extra setup, which should be mentioned in the docu (https://github.com/nodejs/node-gyp#installation)
- node-sass can make problems: `npm rebuild node-sass` can help

## Ideas/Suggestions

- The Interfaces at the moment have a preview Page. Maybe it would be possible to use this functionality and get it all into Storybook, so that it is not redundant
- The storybook build could be build and served as a static page. Could be easy with something like netlify.


.storybook/webpack.config.js
```diff
@@ -0,0 +1,38 @@
+module.exports = storybookBaseConfig => {
```
rijkvanzanten: You're overwriting this module.exports on line 8

.storybook/directusTheme.js
```diff
@@ -0,0 +1,39 @@
...
+  // Form colors
+  // inputBg: 'white',
+  // inputBorder: 'silver',
+  // inputTextColor: 'black',
+  // inputBorderRadius: 4,
+
+  brandTitle: 'Directus',
+  brandUrl: '/',
+  brandImage: 'https://directus.io/img/assets/directus-flat.svg',
+});
```
rijkvanzanten: Can you add a white line to the end of each file? (Git has trouble tracking changes correctly without)

.storybook/manager-head.html
```diff
@@ -0,0 +1,7 @@
+<!-- Storybook override -->
+<script>
+  document.title = 'Directus Components';
+</script>
+
```
rijkvanzanten: Unnecessary whiteline 🤓

src/stories/knobs.stories.js
```diff
@@ -0,0 +1,31 @@
+import { storiesOf } from "@storybook/vue";
```
rijkvanzanten: Is this file in the right location? I thought all the stories would be right next to the component in question, instead of in a special `stories` folder.

henrikwirth: I thought it might be helpful in the beginning to have a stories folder, to test around with it and also there might be stories not related to just one component. Thoughts?

rijkvanzanten: Lets add that when we actually run into the problem of not having a place to store these 'extra stories' instead of preemptively adding it.

src/components/card/card.stories.js
```diff
@@ -6,7 +6,7 @@ import Card from "./card.vue";
 import Readme from "./card.readme.md";
 import { withActions } from "@storybook/addon-actions";
 
-const stories = storiesOf("UI|Card", module);
+const stories = storiesOf("Components|Card", module);
```
rijkvanzanten: There's no spaces between `"Components|Card"`, but there are between `"Components | Button"`

.storybook/directusTheme.js
```diff
@@ -0,0 +1,39 @@
+import { create } from '@storybook/theming';
+
+export default create({
+  base: 'light',
+
+  colorPrimary: "#78909c",
```
rijkvanzanten: Can we replace the hex values here with the Directus global color variables? f.e. `var(--primary)`

.storybook/directusTheme.js
```diff
@@ -0,0 +1,39 @@
...
+  colorSecondary: '#546e7a',
+
+  // UI
+  appBg: '#ebeff1',
+  appContentBg: 'white',
+  // appBorderColor: 'grey',
+  // appBorderRadius: 4,
+
+  // Typography
+  fontBase: '"Roboto", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,\n' +
```
rijkvanzanten: Same variable thing here `var(--family-sans-serif)`

src/components/card/card.stories.js
henrikwirth: Withous whitespace is cleaner, otherwise there will be whitespaces in the sidebar links

rijkvanzanten:
@henrikwirth Wanna check it out? I think this is a workable setup

henrikwirth:
@rijkvanzanten okay, from my side it is done now.

rijkvanzanten:
Does this last change also remove the `notes` tab in the top left? Or will it be visible but empty for every component?

henrikwirth:
> o remove the `notes` tab in the top left? Or will it be visible but empty for every component?



> Does this last change also remove the `notes` tab in the top left? Or will it be visible but empty for every component?

It does remove the note tab completely. So there is only the sidebar or content option through the readme-addon.

henrikwirth:
@rijkvanzanten are we missing anything (besides resolving the conflicts)?

rijkvanzanten:
Don't think so!

PR #1647, opened 2019-05-21
Update table.vue

dev7ch:
remove deprecation warning and move isNil to own method according to new lodash usage.

### DIFF
--- a/src/components/table/table.vue
+++ b/src/components/table/table.vue
@@ -309,7 +309,7 @@ export default {
     }
   },
   methods: {
-    isNil($val) {
+    isNil(val) {
       return _.isNil($val)
     },
     selectAll() {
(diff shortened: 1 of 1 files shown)


src/components/table/table.vue
```diff
@@ -309,6 +309,9 @@ export default {
     }
   },
   methods: {
+    isNil($val) {
```
rijkvanzanten: Why the dollar sign in the name? I thought for a sec that it was a Vue internal..

Lets call it `val` instead of `$val`

```suggestion
    isNil(val) {
```

src/components/table/table.vue
```diff
@@ -309,6 +309,9 @@ export default {
...
+      return _.isNil($val)
```
rijkvanzanten:
```suggestion
      return _.isNil(val)
```

src/components/table/table.vue
dev7ch: ... i thought that about `val` ;)

Issue #1643, opened 2019-05-20
Add "Allow Other" option to Checkboxes


benhaynes:
We need an interface option for `Checkboxes` (CSV, not relational) that (when enabled) allows the user to type another value into a text input. When adding, it would be automatically checked. Then it would be treated the same as any other option... except that if you uncheck it and save it would not longer appear.

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

PR #1653, opened 2019-05-21
Add allow other option to checkboxes

rijkvanzanten:
There are two things left to do before this can be merged:

- [x] Repair the `wrap` option<sup>1</sup>
- [x] Style the custom input

<sup>1</sup> The interface seems to be staging and parsing the value correctly, but for some reason the app is sending the value to the server without the first and last empty `""`. It must be filtered out somewhere. This might not be related to the checkboxes interface itself, but be an issue in the application itself.

Implements #1643


src/interfaces/checkboxes/input.vue
```diff
@@ -65,23 +68,37 @@ export default {
...
+      const optionsWithoutSelection = options.filter(
+        option => this.selection.includes(option.val) === false
+      );
+
+      sortableOptions = [...selected, ...optionsWithoutSelection];
+    } else {
+      sortableOptions = options;
+    }
+
+    // Add a unique ID to each sortable option so we can use that to key the items in the template
```
dev7ch: 👍

PR #1655, opened 2019-05-22
feat: make file-preview usable outside of File Library

### LINKED ISSUE #1229: Make File Preview interface work outside of directus_files

FRSgit:
* create moreOptionsButton component for better managment of optional functionalities popovers
* make file-preview reusable component
* include file-preview as a part of file interface
* unify file-preview's toolbar styling with rest of interfaces
* add label to aspect ratio select
* change `cropper.js` viewMode to overcome problems background adding at photo's edges
* add File Library's response structure (whole values object being a photo) support to file interface
* add imageEditor option to file interface configuration
* add missing translation to file interface configuration
* bump up file interface version

Fixes #1229


FRSgit:
Continuation of #1509 (did new issue, because needed to do things slightly differently because of changes made in directus in the meantime).
Waiting for review!

FRSgit:
Like, I can get only brief change suggestions to my code - it's my first PR here, so I guess there will be some caveats when it comes to code-style and/or logic. Then I can at least work on fixing/changing them ;)

src/components/file-preview.vue
```diff
@@ -99,12 +106,40 @@
 </template>
 
 <script>
-import mixin from "@directus/extension-toolkit/mixins/interface";
+import { i18n } from "../lang/";
```
rijkvanzanten: You don't have to import i18n separately. You can use it's functions directly on the `this` object in Vue, eg `this.$t("edit");`

src/components/file-preview.vue
```diff
@@ -123,21 +158,37 @@ export default {
...
     };
   },
-  watch: {
-    "image.cropRatio"(newValue) {
-      this.setAspectRatio(newValue);
-    }
-  },
   computed: {
+    componentId() {
+      return "file-preview-" + ++id;
```
rijkvanzanten: Instead of using a globally tracked numerical ID, lets rely on shortid instead:

```suggestion
      return "file-preview-" + this.$helpers.shortid.generate();
```

src/components/file-preview.vue
```diff
@@ -123,21 +158,37 @@ export default {
...
+    },
+    moreOptions() {
+      const moreOptions = Object.assign({}, this.options);
```
rijkvanzanten: I've had some troubles where Object.assign sometimes doesn't get rid of the reactivity of the data from Vuex. Relying on lodash's .clone is a safer bet

```suggestion
      const moreOptions = _.clone(this.options);
```

src/components/file-preview.vue
```diff
@@ -123,21 +158,37 @@ export default {
...
+        moreOptions["editor"] = {
+          text: this.$t("edit"),
+          icon: "crop_rotate"
+        };
+      }
+
+      return moreOptions;
+    },
+    titleParsed() {
+      return this.title === null ? this.value.title : this.title;
```
rijkvanzanten: What is the difference between this.title and this.value.title? Where do both come from?

src/components/file-preview.vue
```diff
@@ -123,21 +158,37 @@ export default {
...
+    },
+    selectRatioTranslation() {
+      return i18n.t("select_aspect_ratio");
```
rijkvanzanten: Use `this.$t` instead

```suggestion
      return this.$t("select_aspect_ratio");
```

src/components/more-options-button.vue
```diff
@@ -0,0 +1,90 @@
+<template>
```
rijkvanzanten: This might be very annoying to ask, but would it be possible to make the more-options-button change a separate PR?

Let me know if it's too much work, we can do it at the same time too but it's a bit cleaner to have it separated, seeing that it's a separate change (optimization) from this new feature (enhancement)

src/globals.js
```diff
@@ -92,12 +94,14 @@ Vue.component("invisible-label", InvisibleLabel);
...
 Vue.component("v-avatar", VAvatar);
+Vue.component("v-file-preview", VFilePreview);
 Vue.component("v-card", VCard);
 Vue.component("v-details", VDetails);
 Vue.component("v-notice", VNotice);
 Vue.component("v-table", VTable);
 
 Vue.component("v-upload", VUpload);
 Vue.component("v-progress-ring", VProgressRing);
+Vue.component("v-more-options-button", VMoreOptionsButton);
```
rijkvanzanten: @benhaynes any thoughts on the name `v-more-options-button`? It's the button that's rendered at the right of tables and cards that opens the contextual menu.

Wait, maybe that would be a better name: `v-contextual-menu`

src/interfaces/file/input.vue
```diff
@@ -109,40 +93,50 @@ export default {
     };
   },
   computed: {
+    file() {
+      if (this.value !== null) return this.value;
+      if (this.values !== null && this.values.filesize && this.values.filename) return this.values;
```
rijkvanzanten: Can we add some comments here explaining the difference in flow between `this.value` and `this.values`? I assume `this.values` only applies when this interface is rendered on the detail view of `directus_files`?

src/interfaces/file/input.vue
```diff
@@ -109,40 +93,50 @@ export default {
...
+      return void 0;
```
rijkvanzanten: Prefer `undefined` over `void 0`

```suggestion
      return undefined;
```

rijkvanzanten: > The void operator is often used merely to obtain the undefined primitive value

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void

src/interfaces/file/input.vue
```diff
@@ -109,40 +93,50 @@ export default {
...
-            formatSize(this.value.filesize) +
+            formatSize(this.file.filesize) +
             ")"
         : null;
     },
     src() {
-      return this.value.type && this.value.type.startsWith("image")
-        ? this.value.data.full_url
-        : null;
+      return this.file.type && this.file.type.startsWith("image") ? this.file.data.full_url : null;
```
rijkvanzanten: Lets refactor this into something more verbose and legible, for example:

```suggestion
      const fileType = this.file.type;
      
      if (!fileType) return null;
      
      const isImage = fileType.startsWith("image");
      
      if (isImage) return this.file.data.full_url;
      
      return null;
```

src/interfaces/file/input.vue
```diff
@@ -109,40 +93,50 @@ export default {
...
     },
     icon() {
-      return this.value.type && !this.value.type.startsWith("image")
-        ? getIcon(this.value.type)
-        : null;
+      return this.file.type && !this.file.type.startsWith("image") ? getIcon(this.file.type) : null;
```
rijkvanzanten: Lets refactor this too, see above

src/interfaces/file/input.vue
```diff
@@ -109,40 +93,50 @@ export default {
...
     },
     href() {
-      return this.value.type && this.value.type === "application/pdf"
-        ? this.value.data.full_url
+      return this.file.type && this.file.type === "application/pdf"
```
rijkvanzanten: Lets refactor this too, see above

rijkvanzanten:
This is a really great start for a first PR in a new codebase. Thanks @FRSgit!

I've provided some code-feedback, mostly minor tweaks, no big changes

src/globals.js
benhaynes: I like `v-contextual-menu` :)

src/components/file-preview.vue
FRSgit: Perfect, that one I did in the most straightforward manner, bc I knew that should be better way of handling such cases in this project. Actually I see that in google old directus docs page about `contribution - helpers` appears - why it's not part of current docs anymore?

src/interfaces/file/input.vue
FRSgit: Yup, that's exactly the case and I didn't know if there is better way of handling this (maybe there are some possible changes to directus_files handling itself which I couldn't find?).
I will put comments here

src/interfaces/file/input.vue
FRSgit: Sure, your project, your style guide 😄
Just as a side-noe - minifiers/obfuscators still will convert `undefined` to `void 0`. Also `undefined` can be reassigned to another value in non-global function context, so in theory void <anything> is more secure. But as a matter of fact - it's only about the taste, never saw anybody reassigning `undefined` in real production code 😃

src/components/file-preview.vue
rijkvanzanten: You mean this page? https://docs.directus.io/advanced/app/helpers.html

src/interfaces/file/input.vue
rijkvanzanten: > Just as a side-noe - minifiers/obfuscators still will convert undefined to void 0

Makes sense, saves a couple characters. My main goal is to keep Directus as approachable for external devs as possible; you see `void 0`  way less "in the wild" than `undefined`

> undefined can be reassigned to another value in non-global function context

This is also true, but also plain stupid. It's the same idea as that we could make an object method name a zero-width space ^^

image

src/components/file-preview.vue
FRSgit: o.O seems like google has cached old url for it - https://www.google.pl/search?q=directus+helpers

Nevertheless - from which page helpers docs are accessible? I cannot find a link on contributing page

rijkvanzanten: Oh good question.. I always use the search function 😁 @benhaynes ?

FRSgit: Maybe that's why search crawlers didn't update link as well 😃

src/interfaces/file/input.vue
FRSgit: > plain stupid

Couldn't agree more 😃 
And I get why `undefined` is better for third-party devs, but I blame community for that (imo) bad situation :<
That said I can resolve this conversation ;)

src/interfaces/file/input.vue
FRSgit: Comment will go like this:
```
// needed to overcome incompatibility between directus_files and normal collection detail view
// TODO: investigate why directus_files is still incompatible with normal collection view
```

Is that okay?

src/components/file-preview.vue
FRSgit: `this.value.title` comes from `this.value` prop which is a file object.
`this.title` is a prop value and (probably) was added at some point because there is `subtitle` prop as well. 😄 

Sorry, but it was a while since my first pull request on that and I don't remember everything connected. I must've copied that one now without even noticing that it doesn't make sense at all. I think it can be dropped in favour of using just `this.value.title` (which seems to be present for any type of file).

src/components/more-options-button.vue
FRSgit: I will send additional PR in nex hour or so

FRSgit: One question about this change - from ctx-menu the event is emitted on click - is there any possibility to catch this event in a scope of the same component (so I could just wire up correct handler there)?

E.g. like this (I know that doesn't work):
```
<button ype="button" @edit="something()" @click="$emit('edit')">...</button>
```

src/interfaces/file/input.vue
rijkvanzanten: That's perfect

src/components/file-preview.vue
benhaynes: Fixed! https://docs.directus.io/advanced/app/helpers.html

rijkvanzanten:
Hey @FRSgit! I've updated the branch to use the new contextual menu component. It's looking really good on the file interface, way better than the card we used to have before.

There's one little thing though: the image edit buttons don't seem to do anything.. Not really sure what's causing this as there aren't any issues thrown. Any idea?

rijkvanzanten:
I noticed that it's trying to load in the full image when I enable edit mode too, that might be the cause of the failure:

image

rijkvanzanten:
We also would like to move to a design more similar to this

image

I can take that on if you like once we figure out why the image editing isn't functioning properly

benhaynes:
And here is the empty state:

Screen Shot 2019-05-31 at 5 35 23 PM

FRSgit:
Sorry, had busy week - I tried to investigate issue with CORS error and unfortunately hadnt find bug. I've reverted codebase to commit which was working for sure and well, it doesn't now. I've upgraded API version recently, so to test it I've tried to revert it to state few month before but it hadn't helped.

Obviously there is problem with CORS for 2 requests cropper needs to do to obtain image & exif info about image.

You can face the same problem when you add `crossorigin` attribute to preview's `img` element (line 4 file-preview.vue), e.g.:

```html
      <img v-if="!imgError" ref="image" :src="vUrl" :alt="value.title" @error="imgError = true" crossorigin="anonymous" />
```

The question is if API shouldn't set correct CORS headers for request like that one?

FRSgit:
Ah and - design are beautiful ;) Do you know what will be background color under the image (in situation when image has different resolution that parent container?

benhaynes:
Thanks! The image should be full width and auto height whenever possible. But if it's _smaller_ than that, or is _super_ portrait aspect ratio then we would use `var(--darker-gray)` as the background color and have the image centered.

I defer to @rijkvanzanten or @bjgajjar regarding the CORS issue. :)

rijkvanzanten:
I can't imagine adding CORS in publicly available images will cause any issues... 🤔

creazy231:
Had a CORS problem on another project some days ago. I still don't know why. It seams that there was an update which causes that error but I don't know what has been changed, why and who. Maybe this info helps someone to think in a different right way.

PR #1657, opened 2019-05-22
Fix #1565 – Hide "Add new" and "Select Existing" buttons when field is Read-only

### LINKED ISSUE #1565: Hide "Add new" and "Select Existing" buttons when field is Read-only
Hi, 
I think when a field is Read-only , "Add new" and "Select Existing" buttons should be hidden.

nstCactus:
I based my changes on the files interface

Closes https://github.com/directus/app/issues/1565


src/interfaces/many-to-many/input.vue
```diff
@@ -60,14 +60,16 @@
...
         </div>
       </div>
-      <button type="button" class="style-btn select" @click="addNew = true">
+      <v-button type="button" :disabled="readonly" @click="addNew = true">
         <v-icon name="add" />
         {{ $t("add_new") }}
-      </button>
-      <button type="button" class="style-btn select" @click="selectExisting = true">
+      </v-button>
+      <!--
```
rijkvanzanten: Can we remove this empty comment?

rijkvanzanten:
This is great! Thanks a lot @nstCactus!

src/interfaces/many-to-many/input.vue
nstCactus: I though removing it made the buttons stack rather than being displayed inline but after some more testing I know it doesn't. You can remove it.

benhaynes:
Awesome, really appreciate the PR @nstCactus! 😄

PR #1660, opened 2019-05-22
Fix #1618 – Labels not clickable in field update modal

### LINKED ISSUE #1618: Labels not clickable in field update modal
I'm adding this as a bug because there is a hover event on the label...

Clicking on these labels (field update modal) should toggle the toggle...

Screen Shot 2019-05-13 at 1 20 32 PM

nstCactus:
The `@click.prevent` on div element is there to prevent the `@input` event of the `<input>` to fire twice when clicking on the toggle (clicking the label worked fine without those, but clicking the toggle toggled the value twice, effectively doing nothing).

I also added a focus ring to `v-toggle` components to make keyboard focus visible.

I intentionally didn't squash the commit to let you decide whether you need both changes or just want to keep the bug fix.

Closes #1618


benhaynes:
Thank you @nstCactus!!! ❤️

rijkvanzanten:
I'm a bit confused about the `@click.prevent`. Did clicking on the inner div cause the outer click event to fire twice? 

Should we instead stop propagation of the event by using `@click.stop` on the parent container?

nstCactus:
Exactly, clicking one of the inner divs would fire twice the `input` event. Stoping propagation on the root element of the `v-toggle` component should work as well and is more elegant indeed.

I'll test and amend the PR if this works.

nstCactus:
The real culprit was the `@click` event on the root element of the component: it emitted the value, then the event bubbled to the `<label>`, which changed the `checked` property of the `<input>`, which in turn emitted the value a second time.

rijkvanzanten:
That makes sense. _However_ if the toggle is used standalone (eg without a parent label), now it wouldn't work anymore..

nstCactus:
It should work as long as you have add `v-model` attribute on the component:

````vue
<template>
  …
  <v-toggle v-model="someDataProperty" />
 …
</template>

<script>
  export default {
    data() {
      return {
        someDataProperty: false,
      };
    }
  }
</script>
````

nstCactus:
No I got it wrong, You're right, without a `<label>` the hidden `<input>` cannot receive click events.
I'll fix it again…

nstCactus:
Using @click.stop on the parent container makes the event fire 3 times when the toggle is clicked, and twice when the label is ¯\_(ツ)_/¯.

@click.prevent on the parent works but I fail to understand why.

rijkvanzanten:
> @click.prevent on the parent works but I fail to understand why.

I guess cause it prevents any other action from happening, stopping the event completely? 🤷‍♂ 

(If it works it works!)

rijkvanzanten:
Can we only show the ring when the user is tabbing?

There is a class on the body `user-is-tabbing` whenever the user is using the keyboard to navigate. Lets use that to show the focus ring

rijkvanzanten:
Also, clicking on the toggle directly doesn't put the focus on the input, in turn not showing the outline. Clicking on the label however does put the focus on the input. This is irrelevant if we only show the outline on tab though

rijkvanzanten:
I'll fix that 👍

rijkvanzanten:
Thanks again @nstCactus 😄

nstCactus:
You're welcome. Sorry for the time it took for such a small fix…

rijkvanzanten:
No worries! The bugs that look easy always tend to take up more time.. Glad we figured this out though :)

PR #1666, opened 2019-05-23
Use GH avatar instead of static file

rijkvanzanten:


rijkvanzanten:
Gotta stay on brand 🤷‍♂ 

Who knows, I might switch to that other image everywhere one day

Issue #793, opened 2018-09-08
Facilitate community translations with Crowdin


benhaynes:
#### :rotating_light: MIGRATED FROM REQUESTS.GETDIRECTUS.COM :rotating_light: 

# Facilitate community translations with Crowdin
### :+1: = 1
*Created 3 months ago by @jsamr*
 

-----
### @jsamr – :+1: (3 months ago)
Crowdin allows easy translations from the community, free for OS projects after submitting a request (https://crowdin.com/page/open-source-project-setup-request). I've contributed recently to few projects for french translations and I found it very user friendly and efficient. Would be glad to contribute to this one when you have crowdin setup.

benhaynes:
I have setup an account for this, but it doesn't seem to like our `js` source file... thoughts @jsamr?

Do you think this is the way to go @rijkvanzanten ?

https://crowdin.com/project/directus

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

rijkvanzanten:
I do like Crowdin to have an interface for the translations, that being said, having a single JSON file is also very easy to contribute too 🤔

benhaynes:
I think Crowdin has a few other really nice features — like identifying if keys are missing, auto-translations, etc. I would still be very interested in figuring this out.

benhaynes:
Support for Crowdin has been added and is now live at:

https://locales.directus.io/

Feel free to recommend/contribute your language, we would love the help!! 😄

benhaynes:
We got some great feedback from Crowdin, so I'm adding it here so we don't forget to implement:

-----

Some of the strings are quite difficult to translate without context. I've also checked Russian translations, I guess some strings were translated with lack of context

We have an [In-Context translation tool](https://support.crowdin.com/in-context-localization/). It should work best for a web app. I would recommend using it for at least for most frequently used parts of the UI. [see a demo here](https://demo.crowdin.com/) (I CC our customer care team, we're available 24x7, we'll help to set it up)

We also always recommend adding
 - screenshots wherever possible to provide visual context to translators
 - text description for strings (should easier to add than screenshots)
 - adding proofreaders to review community translations before putting them to production. that might be selected volunteers from the community

it's quite a bit of work but it guarantees the best localization quality possible

PR #1667, opened 2019-05-23
fix: update Polish translations

FRSgit:
updated Polish translations (there were a few missing keys)
sorted en-US translation keys, so it's easier to track missing ones (ofc that's a short-term solution until #793 lands)


rijkvanzanten:
Thanks! Sorting this is a seemingly small change with huge benefits.

FRSgit:
Okay, so I guess that's a green light for sorting rest of lang files, right? Or will you handle it on you end? 😄

rijkvanzanten:
If you could, a PR would be amazing 😄 I'm so busy with everything else that smaller optimizations like these often get ignored..

FRSgit:
Sure, I have one problem with Polish language explicitly which I forgot to speak previously - the Roboto font included in the project doesnt include latin extended typeset because of which Polish is not rendered correctly 100% of a time, see:
image

rijkvanzanten:
Yikes! That should be relatively easy to fix by swapping out the font files in `/assets/fonts`

Could you open an issue for it so we can get it fixed?

rijkvanzanten:
Or a PR of course 😁

FRSgit:
Just sended PR!

PR #1668, opened 2019-05-23
"Interfaces" Interfaces

Nitwel:
Simular to the collections interface.
Just shows all existing interfaces in a dropdown.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "directus",
-  "version": "7.3.0",
+  "version": "7.3.1",
   "private": true,
   "description": "Directus is an Open-Source Headless CMS & API for Managing Custom Databases",
   "author": "Rijk van Zanten <rijk@rngr.org>",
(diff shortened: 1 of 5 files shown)


rijkvanzanten:
Thanks @Nitwel 🙂

benhaynes:
So meta!! 👉 👈

rijkvanzanten:
Can we use the interfaces interface as the interface for an interface option that sets the default interface for the interfaces in the interfaces interface?

Nitwel:
Exactly what my brain was the last few days. 🤣  I had to configure a Json config to develop an Interface that visualizes json configurations which I then could add to the json configuration of the interface.
And the setup of the button groups is also nice because it uses a Repeater field inside a repeater field.

rijkvanzanten:
I remember when creating the interfaces at the beginning, I needed a dropdown interface for a text-input setting, but the dropdown didn't exist yet. I couldn't use a text-input as fallback cause that's the interface I was building 😂

Issue #1462, opened 2019-04-06
2nd level related table getting overwritten


Kinzi:
# Bug Report

Changes on a 2nd level related table cause saving of "root" table in background. This overwrites the password field (I assume hashing the hash again).

## Steps to Reproduce

Let me try to explain better:

1. Set up
I have the following collections and relations:
`Merchant <-- O2M --> Products <-- 02M --> Prices `
The Merchant has the standard hashed password field. Price - Merchant have *no* direct relation

2. Use case causing the error
I am in the products view and open the O2M Modal to edit a price, close the modal and save the product.

3. Error
The change causes the related Merchant to be saved. (as can be seen in it's activity log) After this the hashed value of the password is changed. I assume because the initial hash is hashed again.

More Info:
This does not happen if I make changes only to the product, nor if I make changes to the price directly (without coming from the Products Modal).

I assume it's actually two bugs: O2M modal not saving correctly (this is super buggy anyways https://github.com/directus/app/issues/1412) and the hashed password being rehashed.

## Expected Behavior

Related merchant is not saved at all or at least a hashed password does not get overwritten.

## Actual Behavior

Related merchant is saved and password hash changed.

## Technical Details

- Device: [eg: Desktop, iPhone6, etc]
- OS: [eg: MacOS 10.13.6, Windows 10.1803]
- Browser: [eg: Chrome 68.0.3440.106]
- Directus App: [eg: 7.0.0]
- Directus API: [eg: 2.0.0]
- Install Method: [eg: cloned `master` branch of `directus/directus`]

Kinzi:
As a quick fix should anybody else run into this:

1. Comment out the update hashing in core/interfaces/password/hooks.php (line 33)

2. Write a filter hook for this:

```return [
    'filters' => [
        'item.update.operators:before' => function (\Directus\Hook\Payload $payload) {
            $hashManager = new Directus\Hash\HashManager;

            $password = $payload->get('password');
            // If PW already hashed remove from $payload else hash and overwrite $payload
            if (substr($password, 0, 4) == '$2y$') $payload->remove('password');
            else {
                $hashedString = $hashManager->hash($password, ['hasher' => 'core']);
                $payload->set('password', $hashedString);
            }

            return $payload;
        }
    ]
];```

rijkvanzanten:
The main problem we have to solve first is the deeply nested clones of the parent item, that's preventing updating the M2M now. It's also the cause for many of the O2M bugs floating around.

[...]

PR #1670, opened 2019-05-23
Add ability to sort O2M items

rijkvanzanten:
This PR adds in the ability to manually sort items in a one to many interface. In order to enable this, one has to add a `sort` type field to the related collection.


ThomasKliszowski:
Thanks for this feature @rijkvanzanten 💪

I spotted a tiny bug 🧐
When you add a new item, the `sort` type field is not filled, only on manual dragging.
Is that what you want? This behavior may not be expected when you add an item on an already sorted list.

rijkvanzanten:
Ah good catch! Lemme see what we can do about that

ThomasKliszowski:
I found a more difficult one trying to implement this feature in my use case.
Actually, it seems to be the same _circular reference_ we found with @brocgailit in his **M2M manual sorting implementation** #1500.

It occures when the **Many field** has a `many-to-one` field references himself.
In my use case, I have **A** which has many **B** sorted by your feature on `sort_field`.
But **B** has **translations**, each of these translations references **B**, and repeat the `sort_field`.

When I save, all of my updated `sort_field` fields are ovewritten by the self reference in my translations.

I don't know if I am clear enough, let me know 👍

rijkvanzanten:
That issue was the case with the one to many as well. I was able to solve it for the one to many, as it's easier to retrace what copies of the current item are (it's a direct relationship vs a m2m with a junction table). The best way to fix it is to rework the way the application itself stages values and provides the current edits to the interfaces. That way the m2m can stage only the edits and prevent saving the nested copies.

ThomasKliszowski:
OK got it 👍
This problem seems already tracked by #1462 btw.

PR #1672, opened 2019-05-23
improvement: add VContextualMenu component

FRSgit:
Connected with discussion https://github.com/directus/app/pull/1655#discussion_r287005985

PR #1679, opened 2019-05-25
Fix v-card component: context menu option label was hardcoded to "Remove"

nstCactus:


rijkvanzanten:
The changes made in #1672 fixed this issue as well (albeit in a different way). Thanks anyway @nstCactus! ❤️

PR #1701, opened 2019-05-31
Fix collection prop not passed to layouts

Nitwel:
requires: directus/extension-toolkit#5

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1076,9 +1076,9 @@
       }
     },
     "@directus/extension-toolkit": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/@directus/extension-toolkit/-/extension-toolkit-0.2.0.tgz",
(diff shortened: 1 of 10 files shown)


src/components/card/card.vue
```diff
@@ -51,7 +51,11 @@
           <p v-if="subtitle" class="subtitle">{{ subtitle }}</p>
           <p v-if="body" class="content">{{ body }}</p>
         </div>
-        <v-contextual-menu :disabled="disabled" :options="options"></v-contextual-menu>
+        <v-contextual-menu
+          :disabled="disabled"
+          :options="options"
+          @click="$emit($event)"
```
rijkvanzanten: What is this supposed to do? This doesn't work afaik. `$emit` has two parameters: the event type and the data you're passing on. In this case, it'll result in the event type of `[object Object]`

PR #1706, opened 2019-06-01
Remove warnings

dev7ch:
- remove $lodash warings
- set default status to draft on creation if draft exists
- exclude more data in .gitignore


src/interfaces/status/input.vue
```diff
@@ -59,6 +59,12 @@ export default {
     }
   },
   created() {
+    if (!this.value || this.value === "") {
+      // Set  "Draft" to default if the option exists
+      if (this.$store.state.permissions[this.collection].statuses.draft !== undefined) {
+        this.$emit("input", "draft");
```
rijkvanzanten: We shouldn't hardcode this here for just the name `draft`. The default can theoretically be set to anything

src/routes/file-library.vue
```diff
@@ -278,6 +278,9 @@ export default {
     }
   },
   methods: {
+    keyBy(items, selector) {
```
rijkvanzanten: I think we should be able to access it by adding it in like

```js
methods: {
  keyBy: _.keyBy
}
```

rijkvanzanten:
Thanks @dev7ch! The default for status might be a bit troubling though..

src/interfaces/status/input.vue
dev7ch: yes, its true, but actually i was unsure ttoo about this.

But is there already way so set default status ?

like simply `default: true` ??

src/routes/file-library.vue
dev7ch: yes should work too

src/interfaces/status/input.vue
dev7ch: and i think if this not exist there shouldn`t an error, its like "pseudo hardcoded"  i would say;)

rijkvanzanten: You can set the default for the status interface by using the `default value` field when setting up the interface. We can default the status to the _first_ of the statuses, but we shouldn't do it based on the name of the status, as that's user generated content.

dev7ch: Ok i see

PR #1717, opened 2019-06-05
Fixes #1715

### LINKED ISSUE #1715: WYSIWYG Full: value displayed in tabular view not ideal when the value contains paragraphs
The new WYSIWYG Full interface stores the value in the database as a single line of HTML.
When this value is displayed in tabular view, its HTML tags are stripped. They should be replaced by a space character instead to avoid glueing together the last word of a paragraph to the first word of the next (see the screenshot below).

A quick way to fix this would be to change [display.vue:13](https://github.com/directus/app/blob/c9477f9bccd0be5ab0a5afaaed5f197992edff4d/src/interfaces/wysiwyg-full/display.vue#L13) to replace HTML tags with a space character instead of an empty string (by default multiple whitespace are displayed as a single space character).

Screenshot 2019-06-05 13 19 38

gitlabisbetterthangithub:
Tested and working fine on my dev version of Directus. This is my first time contributing so I may have missed something.

localhost_directus-dev_app_dist_

### DIFF
--- a/src/interfaces/wysiwyg-full/display.vue
+++ b/src/interfaces/wysiwyg-full/display.vue
@@ -10,7 +10,7 @@ export default {
   mixins: [mixin],
   computed: {
     cleanValue() {
-      return this.value ? this.value.replace(/<\/?[^>]+(>|$)/g, "").substring(0, 200) : "";
+      return this.value ? this.value.replace(/<\/?[^>]+(>|$)/g, " ").substring(0, 200) : "";
     }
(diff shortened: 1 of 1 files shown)


benhaynes:
Thank you!

We should make sure this gets applied to the new WYSIWYG editor too @rijkvanzanten

gitlabisbetterthangithub:
New WYSIWYG editor has been fixed in #1718

Issue #1619, opened 2019-05-13
"Sort" should be hidden on detail and browse


benhaynes:
When you create a new collection it has the option to add a `sort` field. When this is selected, the field should be "Hidden on Detail" and "Hidden on Browse" by default since it's mostly only seen as drag handles.

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

benhaynes:
I submitted a PR for this, which is working. However the App is not honoring the `hidden on browse` flag... so I'll open a bug report.

PR #1719, opened 2019-06-05
ref #1619

benhaynes:
also adjusts status draft color, sets a default status


benhaynes:
Just a reminder that this "feature" isn't really done until https://github.com/directus/app/issues/1720 is done

rijkvanzanten:
Nah, this particular change (setting the hidden on browse on by default) is done in terms of it doing what it's supposed to do. #1720 is a bug that's related to this though.

PR #1724, opened 2019-06-06
Interface options interface

rijkvanzanten:
It's the interface options interface to set an interface's interface options. Can be used in conjunction with the interfaces interface in the repeater interface options so we can set the chosen interface's interface options in a dedicated interface options interface.


rijkvanzanten:
Pic or didn't happen:

Screen Recording 2019-06-06 at 3 14 30 PM

shealavington:
image
image

PR #1734, opened 2019-06-07
update form padding and grid template

benhaynes:


src/interfaces/files/input.vue
```diff
@@ -406,8 +406,13 @@ button {
 }
 
 .edit-modal-body {
-  padding: 20px;
+  padding: 30px;
```
rijkvanzanten: 30 is not divisible by 4 and/or 8?

benhaynes: That part isn't my fault, haha — I was just matching the existing padding of the modal header. 

```
.modal-container header[data-v-bc1e7658], .modal-container .footer[data-v-bc1e7658] {
    padding: 20px 30px;
}
```

benhaynes: Obviously eventually the whole app should be updated (and this would be 32px).

rijkvanzanten: Should we update both now? 🤔

benhaynes: We can, but it's also the footer and since the content of other modals is in different places I didn't have time to go through them all to check. Whatever works!

PR #1751, opened 2019-06-12
text should never touch right of hover box

benhaynes:


rijkvanzanten:
What text, what hover box, where does this apply? Before / after?

benhaynes:
There was no padding on the right side of the main nav-bar's background color (the box that appears on hover). So text would flow all the way to the right edge and look "off". The padding was 0px, I added some padding so this doesn't happen.

I try to spend some time each week cleaning up random styles throughout the App, but they are little adjustments across 20 different components. I get the whole accountability thing — but this is the absolutely _most_ I can do for such a simple change. I can't write a paragraph and take screenshots every time I need to adjust an elements padding by 4px. We'll need some sort of short-hand note I can use when a change truly doesn't matter... otherwise I'll end up going back to `style tweaks` PRs with a bunch of random stuff it in.

I like to think I'm relatively aware of what might break things, and/or what needs to be included in release notes. For everything else, I should be able to be a bit more efficient with updates.

rijkvanzanten:
We just need to be able to know what changed without having to go through the code. `text should never touch right of hover box` could apply to nearly everything in the application. Naming it `Add padding on the right side of the main nav-bar's background color` would've been enough already.

We don't need full chapters on the changes for such small tweaks of course

PR #1756, opened 2019-06-13
update highlight color to off-white

benhaynes:
This is not a needed color, so I feel the low contrast is acceptable. But I do like it as a subtle hint.


rijkvanzanten:
acquiesced

PR #1765, opened 2019-06-17
Added checks if config is set correctly

Nitwel:
No more erros when an option is not set.

### DIFF
--- a/src/layouts/timeline/layout.vue
+++ b/src/layouts/timeline/layout.vue
@@ -16,7 +16,6 @@ export default {
     Day
   },
   mixins: [mixin],
-  props: ["items"],
   data() {
     return {
       actionColor: {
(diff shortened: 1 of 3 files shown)


src/layouts/timeline/layout.vue
```diff
@@ -49,22 +48,31 @@ export default {
...
+
+        if (!this.viewOptions.date) return;
 
         var date = new Date(item[this.viewOptions.date].substr(0, 10) + "T00:00:00");
         var existingDay = this.$lodash.find(days, { date: date });
 
-        var color = item[this.viewOptions.color];
-        if (this.fields[this.viewOptions.color].field == "action") {
-          color = this.actionColor[color];
+        var color = null;
```
rijkvanzanten:
```suggestion
        let color = null;
```

src/layouts/timeline/layout.vue
```diff
@@ -49,22 +48,31 @@ export default {
...
+
+          if (
+            this.fields[this.viewOptions.color] &&
+            this.fields[this.viewOptions.color].field == "action"
+          ) {
+            color = this.actionColor[color];
+          }
         }
 
+        var contentType = this.viewOptions.content ? this.fields[this.viewOptions.content] : null;
```
rijkvanzanten:
```suggestion
        const contentType = this.viewOptions.content ? this.fields[this.viewOptions.content] : null;
```

src/layouts/timeline/layout.vue
```diff
@@ -49,22 +48,31 @@ export default {
...
+
         var event = {
```
rijkvanzanten:
```suggestion
        const event = {
```

src/layouts/timeline/layout.vue
Nitwel: Will this change when changing the field in the options?

PR #1775, opened 2019-06-19
Add 2fa login and user setting

OscBacon:
Commits from this PR are needed for the backend to work properly: https://github.com/directus/api/pull/1031

Commit from this PR is need for the sdk to send the otp parameter to the backend: https://github.com/directus/sdk-js/pull/77

This prompts users for their OTP after logging in if they have 2fa enabled. 
It also allows users to enable and disable 2fa in their user settings


benhaynes:
Thanks @OscBacon!!! 😄

OscBacon:
2fa-activation
@benhaynes here's a demo of the 2FA activation workflow :)

benhaynes:
That's excellent! Is that QR code showing after login or upon setup of 2fa? I can't tell because the GIF loops... :)

OscBacon:
Thanks @benhaynes :)
The QR code shows after login if 2fa is enforced but the user hasn't enabled it

benhaynes:
Gotcha. Are they locked to that page until they enable it, or can they delay if they want?

OscBacon:
Locked. The api sends a Missing2FAException otherwise

bjgajjar:
@OscBacon 

Kindly look into the below points.

1. When `Admin` enforces `2FA` for `Admin` role then it is not logging out from the system but it is showing an error. 
     image

2. When user didn't scan QR and then another admin disable 2FA for him then that admin who disables2FA was getting the log out from the system. Let me explain this via an example.    

- Supposed `Admin` role is enforced to enable the `2FA` and we have one admin with an email 
 `admin@example.com` which enabled the `2FA` successfully.
-  Now we have one another admin `admin@rangerstudio.com` which didn't scan QR code and simply press the save button.
-  That user will be able to login to the system once but after logging out user will get the screen of passcode to log in again.
-  As QR was not scanned, the user will not able to find any passcode.
- For this scenario, the user must contact another admin to disable the `2FA`.
- When `admin@example.com` disable the `2FA` for `admin@rangerstudio.com`, the user with  
  `admin@example.com` email will **log out from the system automatically. Which is wrong.**

3. When the system has a single admin who doesn't scan QR and just moving ahead with pressing the "Save" button at that time the user will be in a deadlock and it will become even worst in **`Cloud`** case

@rijkvanzanten Let me know your thought here.

OscBacon:
Hi Binal,
(1) is because the user enforced 2fa for his role, but doesn't have 2fa. This means that on the next page they access, they don't have the right to access the page since 2fa isn't activated but enforced

(3) we can ask the user to enter the OTP under the qr code. Since the only way to have that OTP is to have scanned the qr code, there is a safety net

For (2), could you detail how `admin@example.com` got logged out? I don't see how this could happen

OscBacon:
Also, for (1) @rijkvanzanten any suggestions on how to implement this on the front? I know we had some issues catching a 2fa missing error, and it seems like in this case the api just returns a 300

OscBacon:
Also for (3), what do you think @bjgajjar of having a directus endpoint where you check if an OTP is valid for a given 2fa secret? You'd send both as parameters of the request

bjgajjar:
@OscBacon 

> For (2), could you detail how `admin@example.com` got logged out? I don't see how this could happen

You meant you are not able to reproduce? 

> Also for (3), what do you think @bjgajjar of having a directus endpoint where you check if an OTP is valid for a given 2fa secret? You'd send both as parameters of the request

It seems okay to me.

OscBacon:
@bjgajjar haven't had the time yet, I'm looking to do this in 3 hours or so. That being said, I'm curious as to how (2) could happen. It's a very strange-sounding problem !

bjgajjar:
> It's a very strange-sounding problem !

Agree. Ping me in the slack message if you want any help : )

OscBacon:
> > It's a very strange-sounding problem !
> 
> Agree. Ping me in the slack message if you want any help : )

Sounds good :)

bjgajjar:
This PR seems okay to me. Waiting for #1031 changes to merge it.

rijkvanzanten:
@bjgajjar [You merged a merge conflict](https://github.com/directus/app/commit/74fb6a99b1401b011762c9900ae7c89efa56a77f#diff-32607347f8126e6534ebc7ebaec4853dR14020)

OscBacon:
@rijkvanzanten it's my fault too, I'll make a PR to fix it!

Issue #1759, opened 2019-06-14
Remove support for history routing mode


rijkvanzanten:
It doesn't work combined with using the app in a relative path. Fixing that would add way too much overhead to make it worth it. Ref #1748.

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

dasch8:
I think this could be fixed by using a document base URL (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base). I added the tag manually to `index.html` and it seems to work fine.

benhaynes:
Thanks @dasch8 — we're actually making some big changes to how the App/API are structured... so maybe this won't be an issue after the next release. Any thoughts @rijkvanzanten ?

rijkvanzanten:
@dasch8 We'd have to set the same base in build time to make sure assets like images and script.js / style.css are loaded correctly too, which would break support for running the suite in a folder.

Combining history mode with relative URLs for the static assets doesn't work

Issue #1748, opened 2019-06-12
Field extensions do not load when using history routing mode


gridsystem:
I believe there is a problem with the included .htaccess which routes js and css to the default index.html. You can recreate this problem by switching to history mode in the config, and navigating to a page containing fields.

<img width="1522" alt="Screenshot 2019-06-12 at 4 59 56 pm" src="https://user-images.githubusercontent.com/2365737/59367190-870aef00-8d33-11e9-96a2-d2cb1dbe3474.png">

Inspecting the requests shows that index.html was loaded in place of the expected css or js.

<img width="358" alt="Screenshot 2019-06-11 at 9 46 23 pm" src="https://user-images.githubusercontent.com/2365737/59367125-6773c680-8d33-11e9-9c90-b7e0c33938c0.png">

Switching back to hash routing fixes the issue.

Unfortunately I'm not confident enough with apache configs to submit a PR.

rijkvanzanten:
We can't fix this. These lazy loaded paths are automatically made into absolute URLs by Webpack, which in turn breaks the loading of the interfaces cause the path needs to be relative.

The only real way to fix it would be to provide two different builds for the app, one with relative url support (eg `/admin/index.html`) and one that always lives in the root `/`. This doesn't really solve it fully either, because not everyone who uses history mode wants to have the app in the root level of the domain. 

Seeing that this adds a massive overhead for a feature that doesn't necessarily add any useful features, I'm gonna play [the 80/20 card](https://docs.directus.io/getting-started/contributing.html#_80-20-rule) and consider this not worth it.

rijkvanzanten:
We'll remove support for history mode in the upcoming version. Ref https://github.com/directus/app/issues/1759

gridsystem:
Do you think it's worth adding a line to the docs explaining how to build the vue app from scratch with history mode enabled?

rijkvanzanten:
I'm not sure about that.. If we put it in the docs, we endorse building the application from source, which we don't. I would also require changing some of the buildchain config, basically requiring the user to keep a fork of the application

PR #1777, opened 2019-06-21
Adjusted publicPath settings.

eldarc:
I had the following issue in history routing mode:

In production, CSS and JS resource wouldn't load on routes like `/collections/test`. Since the browser would see them as `/collections/test/css/chunk-0bea28e8.996feaa2.css`, instead of `/css/chunk-0bea28e8.996feaa2.css`.

I made some modification, and this worked for me both in `history` and `hash` mode. It also worked when the app is in a subdirectory.

When the app is deployed in a subdirectory the build needs to be executed as such: `env PUBLIC_PATH=/subd/ npm run build`
Also the `config.js` property `routerBaseUrl` needs to be `/subd/`.

Can someone please see if this issue can be replicated, and if this fixes it. Or if something else is required.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5,14 +5,9 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="shortcut icon" href="<%= webpackConfig.output.publicPath %>favicon.ico">
-    <link rel="manifest" href="/manifest.webmanifest" />
+    <link rel="manifest" href="<%= webpackConfig.output.publicPath %>manifest.webmanifest" />
     <title>Directus</title>
-
-    <% if (process.env.NODE_ENV === "development") { %>
-    <link rel="stylesheet" href="/style.css" />
-    <% } else { %>
-    <link rel="stylesheet" href="./style.css" />
-    <% } %>
+    <link rel="stylesheet" href="<%= webpackConfig.output.publicPath %>style.css" />
   </head>
   <body>
     <noscript>
@@ -21,12 +16,7 @@
     <div id="app"></div>
     <!-- built files will be auto injected -->
 
-    <% if (process.env.NODE_ENV === "development") { %>
-    <script src="/config.js"></script>
-    <script src="/script.js"></script>
-    <% } else { %>
-    <script src="./config.js"></script>
-    <script src="./script.js"></script>
-    <% } %>
+    <script src="<%= webpackConfig.output.publicPath %>config.js"></script>
+    <script src="<%= webpackConfig.output.publicPath %>script.js"></script>
   </body>
 </html>
--- a/vue.config.js
+++ b/vue.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   lintOnSave: false,
-  publicPath: process.env.NODE_ENV === "production" ? "" : "/",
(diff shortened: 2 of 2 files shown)


benhaynes:
Thanks @eldarc!! We'll take a look ASAP... 😊

rijkvanzanten:
Hey @eldarc! Thanks for the update. We'd rather not introduce any build step for the application for installation. As a matter of fact, support for history mode is about to be dropped in favor of having one version that "just works" everywhere with no additional build step involved. See https://github.com/directus/app/issues/1759 and https://github.com/directus/app/issues/1748

eldarc:
Hi @rijkvanzanten 

In my opinion, removing "history mode" isn't a great solution.

I committed a new change to this PR to propose the following:

- With this code, running `npm run build` will build a working version in `hash` mode, without additional configuration required. The app can be deployed in any level directory (relative paths will be used).
- If you want to use history mode, you run `env PUBLIC_PATH=/ npm run build` and history mode will work without any problems when the app is deployed in the root directory.
- If you want history mode in a subdirectory run `env PUBLIC_PATH=/subd/ npm run build` and set the  `routerBaseUrl` to `/subd/` in `config.js`. The app will work from that subdirectory.
- With these changes `yarn dev` or `npm run dev` will work just fine either in `hash` or `history` mode.

If someone doesn't want to mess with routing modes the build will work just fine, but those who want the pretty style can read the docs.

Besides, maybe someone wants to use absolute paths even in hash mode, then they can combine `env PUBLIC_PATH=/ npm run build` with `hash` mode.

For those who need to integrate a custom repo based on `directus-app` in their CI/CD pipelines, they can just add a new script to `package.json`. Something like this: `"build:history": "env PUBLIC_PATH=/ vue-cli-service build",`

This will work for Mac and Linux OSes. For Windows it would look something like this `set PUBLIC_PATH=/ && npm run build`.

rijkvanzanten:
> In my opinion, removing "history mode" isn't a great solution.

Why not? What are the perks of using history mode besides omitting two characters from the URL?

All of these options are great, but make it harder to use (and maintain) the application. I'm still not convinced that all this extra overhead is worth the benefit of having a slightly 'prettier' URL..

Keep in mind that adding this means that people will try using it and fail because of server setup issues in routing, using the wrong path in the command line, trying to build it using an older version of Node, or a plethora of other reasons. These people will all open new issues on this repo and / or reach out on Slack, taking away time and resources for a feature that doesn't add anything as far as I can tell. By having this as an official way of using the app, people also expect us to support it..

eldarc:
I just stated my opinion, since I personally have a use case where I need the history mode. Also, I believe that they will be people who will need it also.

However, if the team thinks this is too much to support, it will be no problem for me to implement this custom code in my mirror repo and merge new versions from the upstream origin.

We can refer to this PR if someone comes up with a similar issue.

There is an apparent use case that is not an issue for most people. But there are clients who will request to remove the `#` from the URLs once you hand-over them the dashboard for use.

rijkvanzanten:
> I just stated my opinion, since I personally have a use case where I need the history mode. Also, I believe that they will be people who will need it also.

What is that use case? Why would one _need_ history mode? Would it enable some functionality that otherwise prevents the usage of the application completely?

I know that certain folks would prefer not having the `#` symbol in the URL, but I'm afraid that by itself isn't enough to justify the added overhead.

PR #1782, opened 2019-06-21
Timeline

benhaynes:
This works best when you also update the default options:

```
Action On
{{ action }}d #{{item}} within {{collection }}
Action By
```
Do those need to be updated in the API? Have they already been added?

Screen Shot 2019-06-21 at 4 49 06 PM


rijkvanzanten:
This changes the avatar in the lower left too. Is that on purpose?

rijkvanzanten:
Those options would only work for en-US though... it won't work i18n wise

PR #1791, opened 2019-06-24
darken icons interface

benhaynes:


rijkvanzanten:
Should we also use off-white for the background? It's more in line with repeater too

benhaynes:
OK, all set.

PR #1792, opened 2019-06-25
Fix some warnings and minor issues with the calendar layout

Kinzi:


benhaynes:
Awesome, thanks @Kinzi! We're back on Directus tomorrow and will check this out!

PR #1817, opened 2019-07-04
Update app.vue

dev7ch:
Adding white bg, coz on scroll you are seeing in image if it uderlayed.

This happens if directus app is installed under, `domain.com/app`.

Update:

www.domain.com <--- a website with a background image
www.domain.com/directus-app <--- directus app with transparent background and > 100vh input form

image

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5,10 +5,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="shortcut icon" href="<%= webpackConfig.output.publicPath %>favicon.ico">
-    <link rel="manifest" href="/manifest.webmanifest" />
+    <link rel="manifest" href="manifest.webmanifest" />
     <title>Directus</title>
 
-    <link rel="stylesheet" href="/style.css" />
+    <link rel="stylesheet" href="style.css" />
   </head>
   <body>
     <noscript>
--- a/src/app.vue
+++ b/src/app.vue
@@ -171,6 +171,9 @@ body.info-wide-active {
 </style>
 
 <style lang="scss" scoped>
+.page-root {
+  background-color: var(--white);
+}
 .error {
   position: fixed;
   top: 0;


benhaynes:
I have no clue where this image is coming from, but the PR makes sense. We'll review and merge ASAP

dev7ch:
Ah, sry, this image comes from my system, its just an example 😉

How it can happen if you install the system as additional app to a existing website.

benhaynes:
> How it can happen if you install the system as additional app to a existing website. 

Haha, now I'm even more confused!! Did you install Directus as an overlay to a site?

dev7ch:
www.domain.com   <--- a website with a bacvkground image
www.domain.com/directus-app <--- directus app with transparent backgrouond¨

Is it clear what i mean?

benhaynes:
But why would the Directus app inherit the background of the root domain? Do they share CSS somehow?

dev7ch:
the image is set via css on `html`  tag

`
html {
    background: url(background.jpg) no-repeat center;
    background-size: cover;
}
`

benhaynes:
Hmm, still weird that Directus and your custom app are sharing CSS... but I'll leave this to @rijkvanzanten to review! :)

dev7ch:
yes, it is, it was not made by me this PR just ensures that directus app is not affected by some server wide messy ccs styles, e.g. on `html` tag.

dev7ch:
@benhaynes Finally the problem seems to occur because directus custom styles are set as absolute path `/styles.css`

image


If there are `/styles.css` the will be loaded, which causes the mixing/shareing of css

src/app.vue
```diff
@@ -171,6 +171,9 @@ body.info-wide-active {
 </style>
 
 <style lang="scss" scoped>
+.page-root {
```
dev7ch: @rijkvanzanten this just should ensure when height is > 100vh there is still a white bg on scroll 🤓

rijkvanzanten: We can get rid of this, as the fix in the other file should fix the root cause of the issue

public/index.html
```diff
@@ -5,10 +5,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="shortcut icon" href="<%= webpackConfig.output.publicPath %>favicon.ico">
-    <link rel="manifest" href="/manifest.webmanifest" />
+    <link rel="manifest" href="manifest.webmanifest" />
     <title>Directus</title>
 
-    <link rel="stylesheet" href="/style.css" />
+    <link rel="stylesheet" href="style.css" />
```
rijkvanzanten: Lets also apply the same change to the custom script file `/script.js`

rijkvanzanten:
Instead of adding the white background specifically, lets just merge the fix for the root problem (the `/style.css` being loaded absolute instead of relative.

public/index.html
dev7ch: it looks like it is already

src/app.vue
dev7ch: yes, if no custom styles are added, but the tiny invisible max-height issue of long forms persists (if you add custom styles you could see it)

sure you wanna drop this "fallback"?

rijkvanzanten: I see what you're saying. Normally I wouldn't mind if something breaks once people start adding custom styles / scripts, but this is an easy enough addition which doesn't hurt things otherwise. We'll leave it in 👍

dev7ch: ok thank you :)

PR #1819, opened 2019-07-05
Turkish translation v0.2 added

anelad:
- The Turkish translation file has some translations added `(*)` to its end. This means that the value is not translated or translated but the translator does not feel right about the translation.
- After reviewing the translation on-the-go these translations will be corrected and maybe some already-translated values will be changed.


rijkvanzanten:
This is great @anelad! Would you like to merge this in as is, or wait until the translations with the `(*)` are improved?

anelad:
@rijkvanzanten I need to check them on-the-go but I do not know how to append the file to my current instance? If there is a way, I would rather check them first. But if there is no way I should merge I suppose.

rijkvanzanten:
@anelad I've added a commit that adds the language to the list of available languages. If you want to check it out locally, you can clone your fork

```
git clone git@github.com:anelad/DirectusApp.git directus-app;
```

install the npm dependencies

```
cd directus-app;
npm install;
```

and run the app 

```
npm run dev;
```

that should fire it up for debugging on localhost:8080

anelad:
@rijkvanzanten Thanks, man. I'll be on it tomorrow.

anelad:
It looks like `user profile page`s are not translatable for now.

anelad:
Turkish translation is ready for public usage. You can merge it @rijkvanzanten

benhaynes:
Thanks @anelad! This should make it into today's release!

autr:
Hi, this translation seems is causing syntax errors for me:

```
 warning  in ./src/interfaces/collections/meta.json

Module parse failed: Unexpected token } in JSON at position 1749 while parsing near '...ahil eder"
    },
  }
}
'
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected token } in JSON at position 1749 while parsing near '...ahil eder"
    },
  }
}
```

I'm using:

npm 6.10.1
node v10.14.2

rijkvanzanten:
Interesting, I just bundled everything in production and didn't see this issue!

rijkvanzanten:
I do see that trailing comma in the file though. Wondering why it didn't cause problems on my end 🤔

autr:
Another trailing commas in calendar/meta.json - removed both and working fine :)

rijkvanzanten:
Good catch. I've just pushed a fix. Weird that that fails silently in Vue CLI! Oh well 🤷‍♂

anelad:
Translations of interfaces not planned for this pull, my bad not separating them.

Issue #1822, opened 2019-07-06
Can't upload image | Uploading image freezes UI


edenprojectde:
When i try to upload an image i get a vue error about the filename/spliting.
The interface gets stuck and is non-interactable the moment i selected a file.

OS: Linux MX 18
Browser: Chrome
Server Stack: NGINX / PHP 7.3
Directus: v7.6.1

![image](https://user-images.githubusercontent.com/6115966/60749551-c4157a80-9f9b-11e9-85d2-9868bdab9d56.png)

- The upload doesn't seem to fail, the file is on the Server and i can select it when using the Button "Existing" but no thumbnail is being genereated.
- The Thumbnail seems to fail to be generated when this error occurs.
  - Fixed the whole Thumbnail Problem with this Nginx Setup: https://docs.directus.io/advanced/server-setup.html#nginx

rijkvanzanten:
Fixed in https://github.com/directus/app/pull/1832

PR #1832, opened 2019-07-09
Fix issue #1822 uploading image freezes UI

nestebe:


benhaynes:
Thanks @nestebe!! We'll review asap... 😄

PR #1848, opened 2019-07-15
Fix 1573 - File duplicates in files interface

eldarc:
- The `selectItems` method removed the junction row id for items, so the patch option didn't know what items to update. This resulted that some files were duplicated.
- After the fix was applied the delete failed silently if the `$delete` items are not at the end of the array. Added sort to `deleteFile`.


src/interfaces/files/input.vue
```diff
@@ -299,34 +298,47 @@ export default {
...
+              ...junctionRow,
+              $delete: true
+            };
+          }
 
-        return junctionRow;
-      });
+          return junctionRow;
+        })
+        // Items with `$delete` should go to the end of the array to be properly deleted. TODO: Check if this is a bug
```
rijkvanzanten: This definitely sounds like a bug! Is this reproducible in just the API?

eldarc: Tested the API through Postman and got the same behavior. It appears to be a bug.

rijkvanzanten: @eldarc would you mind opening an issue for that on the API repo? 🙂

eldarc: @rijkvanzanten 

I investigated a bit more and found out that the order is not the problem. The issue occurs because the `junctionRow` has the `[collection]_id` property. In there is a reference to all fields of the current item. However only the initial value is stored, and if there any updates to the fields they are not reflected here.

The patch method somehow reads the latest data (end of the array) from the `files` field instead of using the root properties. That's why the delete issues occurred.

The solution is just to leave the `id` property of the `[collection]_id`  object.

shealavington: I've tested this fix, and it works for me, however, @eldarc, are you saying that this issue is entirely frontend dependant and that this isn't an API issue, or is it also an API issue?

eldarc: @shealavington Depends.

Does the API never expect such circular data structure? If not, then it's purely front-end.

shealavington: cc: @rijkvanzanten

rijkvanzanten: The application shouldn't be sending circular copies of the parent item. Ideally, it only sends the values that were changed, but that will probably take more refactoring work to implement.

eldarc: Can we then maybe think about testing and merging this fix until there is an initiative to change the behavior by refactoring?

With the latest changes, it doesn't use `sort`, it just removes all the outdated circular data.

rijkvanzanten:
Ref: 8c5f725209bbbc9c9e6e3cafe40f88fc38c98013

PR #1868, opened 2019-07-20
Update items.vue

benydc:
Populate fields selected from view query.

This makes the request faster allowing only certain fields to populate using the view query.

Fixes https://github.com/directus/app/issues/1871

### DIFF
--- a/src/components/items.vue
+++ b/src/components/items.vue
@@ -392,8 +392,10 @@ export default {
 
       Object.assign(params, this.viewQuery);
 
-      // The above viewquery can override the fields. We want to force load all the fields so we don't end up with missing crucial data
-      params.fields = "*.*";
+      if (this.viewQuery && this.viewQuery.fields)
+        params.fields = params.fields.split(',').map(field => `${field}.*`).join(',');
+      else
+        params.field = "*.*";
 
       if (this.searchQuery) {
         params.q = this.searchQuery;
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @benydc! We'll review asap! ❤️

rijkvanzanten:
@benhaynes this does make your plan for the progress interface impossible. IMO that's a tradeoff that's worth the performance gain on this page. Thoughts?

rijkvanzanten:
Ref: there was a plan to be able to show a progress bar on this page based on what values where filled out in the row, but we would need to fetch all the columns in the row to be able to do that.

benhaynes:
Performance is the name of the game! Couldn't we do a "progress" interface asynchronously if we really wanted/needed it?

src/components/items.vue
```diff
@@ -392,8 +392,14 @@ export default {
...
 
-      // The above viewquery can override the fields. We want to force load all the fields so we don't end up with missing crucial data
-      params.fields = "*.*";
+      if (this.viewQuery && this.viewQuery.fields) {
+        params.fields = params.fields
+          .split(",")
+          .map(field => `${field}.*`)
+          .join(",");
+
+        if (!params.fields.includes("id")) params.fields += ",id";
```
rijkvanzanten: We can't rely on the hardcoded name `id` here, as the primary key can be called anything.

PR #1881, opened 2019-07-24
Reimagine files interface

rijkvanzanten:
It now uses a similar structure as the many to many interface, including the ability to sort items.


rijkvanzanten:
image

cdwmhcc:
sort is great!
No longer need to delete and re-select

cdwmhcc:
@rijkvanzanten 

image

Files are not sorted in my actual use.
Why are files sortable in this diagram

rijkvanzanten:
If the junction table has a field with the `sort` type, the interface will allow you to manually sort

cdwmhcc:
> If the junction table has a field with the `sort` type, the interface will allow you to manually sort

OK,Thank you

Issue #1874, opened 2019-07-19
Fields in translation are not update


jbitsch:
# Bug Report

## Steps to Reproduce

1. git clone `https://github.com/directus/api .` and setup scheme based on src/scheme.sql
2. `composer install`
3. Copi version of api.php
4. Create 3 new collections called `test_main` `test_translation` and `language`
5. The `language` collection contains the field `name` and a primary key `id`. 
6. Insert one language entry in the `language` collection
7. Create the two fields `test_text` and `language_code` in the `test_translation` collection
8. Create a translation field in the `test_main` which links to the  `test_translation` collection:
![image](https://user-images.githubusercontent.com/1073322/61548961-32077a80-aa4f-11e9-85ed-f130218778f5.png)

9. Create an entry in the `test_main` collection and see that i saves the translation correctly. 
10. Try to update the text in the translation.

## Expected Behavior
When creating an item the fields in the translation should be saved. 
When updating an item the fields in the translation should be updated.

## Actual Behavior
When creating an item the fields in the translation should be saved.  (It does)
When updating an item the fields in the translation are NOT being updated.

## Technical Details

- Device: Desktop
- OS: MacOS + Google App Engine
- Web Server: Apache 2.4.37
- PHP Version:  7.2.0
- Database:  MySQL 8.0.12
- Install Method:  cloned `master` branch
- Use the master branch for both the app and the api

Issue #1849, opened 2019-07-15
Second level translated fields are stored but not rendered back on read


chintohere:
### Versions 
App: 7.6.2
Api: 2.2.1

### How to  reproduce
Create collection structure with second level translated fields in directus.

[schema diagram](https://dbdiagram.io/d/5d2c4d6bced98361d6dca889)

1. Create new article
2. Add new section
3. Add translated value *section_name*
4. Save. Observe that save is successful
5. Read the article again. Notice that the translated fields are empty

oliverpool:
We are also affected by this bug.

Looking at the Network requests, it appears that the data as been saved (is present in the response under the `translations` object).

I have some coding knowledge and I am going to dig further (I guess the issue is in the `Translation` component, since the API response looks fine).

Issue #1150, opened 2018-11-27
Error when adding a translation field


sebj54:
# Bug Report

## Steps to Reproduce

1. Create a collection `test`
2. Create a translation collection `test_translations` (with language_code, foreign key and a test text field)
3. Add a translation field to test collection

⚠️ I'm not sure it is the actual reproduction scenario. If you can't reproduce, let me know, I'll provide screenshots. It worked before (2-3 hours ago) as I actually have a working translation collection. Really weird bug!

EDIT: Apparently, the "relation" configuration of the field is not saved on creation. The translation field is correctly added to the table because if I reload the page, I can see it. When I edit the field, I can edit the relation to configure it correctly. It works after that!

## Expected Behavior

Translation field should be added

## Actual Behavior

Error

## Other Context & Screenshots

POST request: http://local.directus.com/_/relations
Payload:
```
[...]

PR #1902, opened 2019-08-01
Translation interface fields blank fix

JackLeEmmerdeur:
Referencing [#1874](https://github.com/directus/app/issues/1874), [#1849](https://github.com/directus/app/issues/1849), [#1094](https://github.com/directus/app/issues/1094), [#1150](https://github.com/directus/app/issues/1150), [#1231](https://github.com/directus/app/issues/1231) and [#directus/1766](https://github.com/directus/directus/issues/1766)

Inside a translation-interface-field the contents of the translated fields are empty, although they are saved to the database translation-relations table.  

In the current state the second argument to the _.keyBy function will always result to "language.code". The resulting object from _.keyBy will return an object with one item with an undefined key and a translation-item.

According to  [_.keyBy](https://lodash.com/docs/4.17.15#keyBy) the second parameter is defined as an iteratee-function.

This fix will generate correct values for the :value parameter in the v-form eventually.

BTW:
I might have a fix for [#directus/1215](https://github.com/directus/directus/issues/1215), too.
In my case if I had more than one translateable field in a translate-interface, the app mixed the values up. Fix is in the same file: "input.vue". How should I proceed there? 

- Create a new branch (I guess). 
- Commit to the current branch (translation-interface-fields-blank-fix), push and create a new pull request.


rijkvanzanten:
https://github.com/directus/directus/issues/1215 Is from a previous major version (totally different codebase), so even if the symptoms might be similar, it can't be the same bug. Anyway, a separate PR for the other fix is appreciated :)

JackLeEmmerdeur:
Hey @rijkvanzanten. Regarding your comment, I misread the issue text in #direcuts/1215 and will set that aside for now.

Now to my merged fix: I believe my commit was missing `if (_.get(translation, this.options.translationLanguageField) === this.activeLanguage)` in now line 146 of src/interfaces/translation/input.vue.

Could you some merge the code of pull request [#1914](https://github.com/directus/app/pull/1914) from oliverpool, or does it conflict with my commit? If yes, just overwrite my changes, as oliverpools changes work on my side.

PR #1879, opened 2019-07-24
Attempt to fix rendering of translated fields

oliverpool:
Fix #1849 

The `valuesByLang` was:

```
[
  {
    description: "some description",
    language: { language: "en" }
  }
]
```

but it should be:

```
[
  {
    description: "some description",
    language: "en"
  }
]
```

PR #1914, opened 2019-08-01
Fix regression of #1879

oliverpool:
This should allow the translation module to work if the related fields are loaded or not.

See https://github.com/directus/app/pull/1879#issuecomment-516383888


oliverpool:
Testable here: https://deploy-preview-1914--directus-nightly.netlify.com/

PR #1915, opened 2019-08-02
Proposal: a change to the language used for single-file field's options

justrealmilk:
At the moment, once a file is selected, the way to "remove" it is to open the 3 dot options menu and select "Delete." As this doesn't delete the file itself, I have a proposal in order to make it more clear to users.

image

Changes icon and term.


src/interfaces/file/input.vue
```diff
@@ -10,8 +10,8 @@
       :href="href"
       :options="{
         remove: {
-          text: $t('delete'),
-          icon: 'delete'
+          text: $t('Deselect'),
```
rijkvanzanten: Lets make the key a lowercase `deselect` and also make sure to add this string to the languages file so it can be translated

`/lang/en-US/index.json`

rijkvanzanten:
Alternatively, we could also consider adding an interface option that allows you to toggle between "deselect" and "delete" so the interface will actually fully delete the related image.

benhaynes:
Couldn't we offer both options? Deselect file and delete file?

rijkvanzanten:
Fair enough!

src/interfaces/file/input.vue
justrealmilk: super dooper trooper

justrealmilk:
As requested! 

image

benhaynes:
Awesome!! Thank you @justrealmilk! 😄

rijkvanzanten:
> Remove continues to double up as deselect as I'm unsure at this point how to do the thing

I did the thing for ya 😄

Go team!

PR #1919, opened 2019-08-02
revert field note position and update field label styling

benhaynes:


rijkvanzanten:
Hover effect / clickable zone of what?

benhaynes:
Changes the required color to `accent`, changes the hover of field action to `accent`, and increases the field action clickable area to include the field name.

benhaynes:
Just a head's up that this is now a bigger change, and revert the Field Note back to how it was before. Eg: Icon gone, inline gray markdown is back.

Screen Shot 2019-08-03 at 12 13 43 PM

benhaynes:
@rijkvanzanten any reason this wasn't merged?

PR #1941, opened 2019-08-09
Fix #1278

### LINKED ISSUE #1278: Incorrect data when switching between projects
# Bug Report
When you login to a different project on the same app instance in the same browser window, you still see cached data from the latest project (collections, user name, etc.), even though you should already see the new project. Reloading the window helps to fetch the correct data.

## Steps to Reproduce

1. Login to some Directus project
2. Logout
3. Select a different project with different collections, so you can see the difference and login again
4. In the side menu, you should now see collections from the first project.
5. Reload the window and the new collections appear/old collections disappear

## Technical Details
- Directus App: [eg: 7.0.15]
- Directus API: [eg: 2.0.15]

theharshin:
```js
const newState = Object.assign({},initialStateCopy);
```
Somehow here `newState` was keep referencing the `initialStateCopy`. Changed that to `JSON.parse(JSON.stringify(initialStateCopy))` and all working fine! 🤔

@rijkvanzanten  Any idea why such behavior?

Fixes #1278

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -50,7 +50,7 @@ export function resetState() {
   // the store.replaceState method will make the passed in object reactive.
   // This will make a clone to modify, so the original initialStateCopy stays
   // as-is
-  const newState = Object.assign({}, initialStateCopy);
+  const newState = JSON.parse(JSON.stringify(initialStateCopy));
   newState.auth.projectName = store.state.auth.projectName;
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
I guess object assign doesn't get rid of the Vue reactive prototype methods 🤷‍♂ 

I've been using `_.clone()` instead of this method everywhere else 👍

PR #1945, opened 2019-08-09
Fix #1742

### LINKED ISSUE #1742: SSO - Last Activity Page
When you have SSO Enabled ( or maybe even without ) 
/login  should never be in your  "last_page"  visited user profile field. 

It took me a few hours to figure out why my SSO didn't work anymore.

After a succesfull login your see a http query:
me?fields=last_page   -->    {"data":{"last_page":"\/login"}}

which caused the "app"  to show the login page => I never was able to go to the app, because I got stuck on the login page. 

how to reproduce?
put /login   in the database as the last loggedin page => you have your issue.

how to solve?
- or make sure it never reaches the DB. '/login'
- or if '/login' is detected from client side . bypass to go to the dashboard

theharshin:
Fixes #1742

### DIFF
--- a/src/routes/login.vue
+++ b/src/routes/login.vue
@@ -4,11 +4,9 @@
       <v-install v-if="installing" :saving="saving" @install="install" />
 
       <form v-else @submit.prevent="processForm">
-        <img class="logo" alt="" src="../assets/logo-dark.svg" />
+        <img class="logo" alt src="../assets/logo-dark.svg" />
 
-        <h1 v-if="loading">
-          {{ loggedIn ? $t("fetching_data") : $t("signing_in") }}
-        </h1>
+        <h1 v-if="loading">{{ loggedIn ? $t("fetching_data") : $t("signing_in") }}</h1>
         <h1 v-else-if="notInstalled">{{ $t("welcome") }}</h1>
         <h1 v-else>{{ resetMode ? $t("reset_password") : $t("sign_in") }}</h1>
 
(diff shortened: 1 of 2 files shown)


src/routes/login.vue
```diff
@@ -4,11 +4,9 @@
       <v-install v-if="installing" :saving="saving" @install="install" />
 
       <form v-else @submit.prevent="processForm">
-        <img class="logo" alt="" src="../assets/logo-dark.svg" />
+        <img class="logo" alt src="../assets/logo-dark.svg" />
```
rijkvanzanten: `alt` should have a value of `""` (for a11y reasons)

rijkvanzanten:
@bjgajjar @OscBacon could you guys checkout the conflict in `src/router.js`? The 2FA merge updated the same bit of code and I want to make sure we don't remove anything on accident that's needed for that

OscBacon:
@theharshin @rijkvanzanten I can't contribute to the branch but here's the fix I suggest:
```
-  if (store.state.hydrated && from.path !== "/logout" && from.path !== "/login") {
+  // Prevent tracking if not logged in
+  if (to.path === "/2fa-activation" && api.loggedIn === false) {
+    return;
+  }
+
+  if ((store.state.hydrated && from.path !== "/logout" && from.path !== "/login") || to.path !== "/2fa-activation") {
```

theharshin:
Updated the PR 🙂

PR #1963, opened 2019-08-14
Fix issue with params.fields as Array

benydc:
params.fields can be Array type, when using the layout modal for many-to-one.


rijkvanzanten:
Thanks @benydc! Does this close any specific issue (that has a separate ticket on GH)?

benydc:
> Thanks @benydc! Does this close any specific issue (that has a separate ticket on GH)?

@rijkvanzanten nope, but we encountered this issue.

rijkvanzanten:
No problem! Just making sure we don't leave any closed issues open on the bugtracker :)

PR #1971, opened 2019-08-17
Preserve selected file info on change of language

itsmerhp:
**Issue:** On change of language fileinfo is not preserving.
**Solution:** Preserved file details in **file_data** attribute to resolve the issue.

### DIFF
--- a/src/interfaces/file/input.vue
+++ b/src/interfaces/file/input.vue
@@ -199,7 +199,7 @@ export default {
     saveUpload(fileInfo) {
       this.image = fileInfo.data;
       // We know that the primary key of directus_files is called `id`
-      this.$emit("input", { id: fileInfo.data.id });
+      this.$emit("input", { id: fileInfo.data.id, file_data: this.image });
 
       this.newFile = false;
     },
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
⚠️  This is not a good solution and needs to be reverted @bjgajjar @itsmerhp @directus/team 

The solution here is _not_ to stage a custom arbitrary `file_data` key that's checked by a hardcoded reference in the translations interface. It's a very hacky exception that only works for this very specific use case. 

The underlying issue here seems to be that the file interface isn't able to render the image based on the content that's provided coming in from the translations interface. We should investigate why this is the case and fix that issue instead.

rijkvanzanten:
Secondly, this kind of code is very illegible and should be avoided:

```js
value = value !== null && value.file_data !== undefined ? value.file_data : value;
```

Instead, consider using a more legible if statement:

```js
if (value && value.file_data) {
  value = value.file_data;
}
```

cdwmhcc:
> Secondly, this kind of code is very illegible and should be avoided:
> 
> ```js
> value = value !== null && value.file_data !== undefined ? value.file_data : value;
> ```

Yeah, it's a complicated

PR #1973, opened 2019-08-19
Fix #1966 [Resolve : Card layout displays [object Object]]

### LINKED ISSUE #1966: Card layout displays [object Object] while table layout output is correct
When selecting cards layout and setting title or subtitle to an O2M field, the output is [object Object] while everything is displayed correctly in the table layout.

App version 7.9.0
API version 2.4.0
PHP 7.3
MariaDB 10.3.11
Apache 2.4.29

bjgajjar:
@theharshin — This issue can be solved with using `slot` as right now it is rendering via `props`.

I already used `slot` in this PR and for that, the dynamic render of extension must be there as it should be changed as per the value selected in view option(info bar).

Right now extensions are not being loaded at a run time and throwing an error when the value is getting changed from view option. You can find that error below.

image

Kindly look into it so we can merge this PR.

### DIFF
--- a/src/components/card/card.vue
+++ b/src/components/card/card.vue
@@ -46,11 +46,17 @@
       </div>
       <div class="body" :class="{ menu: options != null }">
         <div class="main">
-          <component :is="titleElement" v-tooltip="title" class="title">
-            {{ title }}
-          </component>
+          <slot name="title">
+            <component :is="titleElement" v-tooltip="title" class="title">
+              {{ title }}
+            </component>
+          </slot>
+          <!-- <slot name="subtitle"> -->
           <p v-if="subtitle" class="subtitle">{{ subtitle }}</p>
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
The ext-display needs to be used for all three possible positions: title, subtitle, and body.

rijkvanzanten:
> Right now extensions are not being loaded at a run time and throwing an error when the value is getting changed from view option.

This is in this PR specifically. I think this can be fixed by adding a `key` prop to the `v-ext-display` which matches the name of the field that's being rendered. If the field name changes, the key will be updated which in turn re-renders the v-ext-display, which will reinitialize it and thus fetch the interface fresh

theharshin:
Closing this in favor of https://github.com/directus/app/pull/2001

PR #1981, opened 2019-08-22
Fix merge conflict in package-lock

OscBacon:
Fix merge conflict introduced in https://github.com/directus/app/pull/1775


OscBacon:
@bjgajjar @rijkvanzanten 
I merged the conflict by keeping both packages.
That being said, do we know if both packages are used ? 
(pnp-webpack-plugin and pngjs)

PR #1997, opened 2019-08-29
Fix #1990 [Sorting Preview interface issue]

### LINKED ISSUE #1990: Preview Interface is sortable on browse
Hi folks, I really love Directus!
There is a caveat while sorting via a collections browser list, that should be addressed.
I have a collection, using a "preview"-interface. Even though I've set "hidden on browse", my current version seems to display it anyway. This should be fixed with: https://github.com/directus/app/issues/1720

But for now, I _can_ sort it, but than my whole view is broken, telling me a "Server error has occured". My API backend is logging the following:

```
[2019-08-28 09:30:24] api[_].ERROR: Directus\Database\Exception\UnableSortAliasFieldException: Unable to sort by "product_preview" field because it is an alias. in /var/www/html/src/core/Directus/Database/TableGateway/RelationalTableGateway.php:1573
Stack trace:
#0 [internal function]: Directus\Database\TableGateway\RelationalTableGateway->processSort
#1 /var/www/html/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(799): call_user_func_array
#2 /var/www/html/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(1012): Directus\Database\TableGateway\RelationalTableGateway->applyParamsToTableEntriesSelect
#3 /var/www/html/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(1137): Directus\Database\TableGateway\RelationalTableGateway->fetchItems
#4 /var/www/html/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(819): Directus\Database\TableGateway\RelationalTableGateway->fetchData
#5 [internal function]: Directus\Database\TableGateway\RelationalTableGateway->getItems
```

Since the user was not able to repair this on his own, the only way to fix this was reset the user's collection preset, by `mysql> DELETE FROM directus_collection_presets WHERE id = 27 AND user = 26 LIMIT 1;`

Is there a flag for any Interface to stop allowing "sorting" on items browse list?

----

My versions:

* APP: v7.0.18 (Docker)
* API: v2.0.18 (Docker)
* MySQL: v5.7

urvashithakar:
When sorting `ALIAS` type like `Preview` interface, "Server error has occured" in the table layout of APP.
This PR solved this issue.


rijkvanzanten:
This is not a good solution. The app can still sort on preview this way when another layout is used. There's already a top level filter for relational tiers in items.vue that should be extended.

@bjgajjar can you please revert this?

bjgajjar:
Okay, so what's the best solution for this? How can we implement the sorting on `Alias` field? Can you please provide your guidance?

rijkvanzanten:
This solutions will prevent the app from sorting _after_ the user clicks on the column. The column will still have the hover state + arrow show up.

Here:

https://github.com/directus/app/blob/076baea74de5b08f7dd1c917a08e9e64db62aae9/src/components/table/table.vue#L30-L32

and here:

https://github.com/directus/app/blob/076baea74de5b08f7dd1c917a08e9e64db62aae9/src/components/table/table.vue#L46-L51


is where we want to add the check for `alias` as well

PR #2014, opened 2019-09-02
Fix/1993 [Preview Field URL Doesn't Contain Variable in List View]

urvashithakar:
In `Table Layout` and `Card layout`, when we select `Preview` interface it is showing -- in listing or card. Also in `M2M` and `O2M` mapping table listing when displaying `Preview` field, URL doesn't contain mentioned variable. This is fixed in this PR.

### DIFF
--- a/src/components/item-select/item-select.vue
+++ b/src/components/item-select/item-select.vue
@@ -94,6 +94,7 @@
               :datatype="fieldInfo.datatype"
               :options="fieldInfo.options"
               :value="item[fieldInfo.field]"
+              :values="getItemValueByID(item.id)"
             />
           </span>
         </label>
@@ -284,6 +285,14 @@ export default {
         params.fields = []; // ISSUE#1865 Fixed Define the blank fields array to push the data.
       }
 
+      const fieldsData = this.fields.map(fieldName => {
+        return this.$store.state.collections[this.collection].fields[fieldName];
+      });
+
+      // ISSUE#1993 Preview Field URL Doesn't Contain Variable in List
+      var result = fieldsData.filter(field => field.type.toLowerCase() === "alias");
+      if (result.length > 0) params.fields = ["*.*"];
+
       let sortString = "";
       if (this.sortDirection === "desc") sortString += "-";
       if (this.sortField) sortString += this.sortField;
@@ -346,6 +355,10 @@ export default {
         offset: offset,
         replace: false
       });
+    },
+    getItemValueByID(id) {
+      var value = this.items.filter(item => item.id == id);
+      return Object.assign({}, value[0]);
     }
   }
 };
--- a/src/components/item-select/item-select.vue
+++ b/src/components/item-select/item-select.vue
@@ -94,7 +94,7 @@
               :datatype="fieldInfo.datatype"
               :options="fieldInfo.options"
               :value="item[fieldInfo.field]"
-              :values="getItemValueByID(item.id)"
+              :values="getItemValueById(item.id)"
             />
           </span>
         </label>
@@ -285,11 +285,10 @@ export default {
         params.fields = []; // ISSUE#1865 Fixed Define the blank fields array to push the data.
       }
 
+      // ISSUE#1993 Preview Field URL Doesn't Contain Variable in List
       const fieldsData = this.fields.map(fieldName => {
         return this.$store.state.collections[this.collection].fields[fieldName];
       });
-
-      // ISSUE#1993 Preview Field URL Doesn't Contain Variable in List
       var result = fieldsData.filter(field => field.type.toLowerCase() === "alias");
       if (result.length > 0) params.fields = ["*.*"];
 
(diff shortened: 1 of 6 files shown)


src/components/item-select/item-select.vue
```diff
@@ -346,6 +354,10 @@ export default {
         offset: offset,
         replace: false
       });
+    },
+    getItemValueById(id) {
+      var value = this.items.filter(item => item.id == id);
```
rijkvanzanten:
```suggestion
      const value = this.items.filter(item => item.id == id)[0];
```

src/components/item-select/item-select.vue
```diff
@@ -346,6 +354,10 @@ export default {
...
+      return Object.assign({}, value[0]);
```
rijkvanzanten:
```suggestion
      return Object.assign({}, value);
```

src/components/table/table.vue
```diff
@@ -116,11 +116,19 @@
...
+                    fieldInfo.type.toLowerCase() != 'alias'
+                "
+                class="empty"
+              >
                 --
               </div>
               <v-ext-display
-                v-else-if="useInterfaces && !isNil(row[field])"
+                v-else-if="
+                  useInterfaces && (!isNil(row[field]) || fieldInfo.type.toLowerCase() == 'alias')
```
rijkvanzanten:
```suggestion
                  useInterfaces && (!isNil(row[field]) || fieldInfo.type.toLowerCase() === "alias")
```

src/components/table/table.vue
```diff
@@ -116,11 +116,19 @@
               }"
               class="cell"
             >
-              <div v-if="row[field] === '' || isNil(row[field])" class="empty">
+              <div
+                v-if="
+                  (row[field] === '' || isNil(row[field])) &&
+                    fieldInfo.type.toLowerCase() != 'alias'
```
rijkvanzanten:
```suggestion
                    fieldInfo.type.toLowerCase() !== "alias"
```

src/components/items.vue
```diff
@@ -407,6 +406,11 @@ export default {
         params.fields = "*.*";
       }
 
+      // ISSUE#1993 Preview Field URL Doesn't Contain Variable in List
+      const fieldValues = Object.values(this.fields);
+      var result = fieldValues.filter(field => field.type.toLowerCase() === "alias");
```
rijkvanzanten:
```suggestion
      const result = fieldValues.filter(field => field.type.toLowerCase() === "alias");
```

src/components/items.vue
```diff
@@ -407,6 +406,11 @@ export default {
...
+      if (result.length > 0) params.fields = "*.*";
```
rijkvanzanten: This will fetch all nested fields in all top level fields as soon as there is 1 field that is an alias. That's pretty hefty performance regression. If possible, we should only fetch `.*` on the actual `alias` type fields.

We might be able to do this by generating `params.fields` based on fieldValues

For example:

```js
let params = "*";

fieldValues
  .filter(field => field.type.toLowerCase() === "alias")
  .forEach(field => {
    params += `,${field.field}.*`;
  });
```

This should make sure that `fields` will end up like `*,preview_field.*,other_alias_field.*`

urvashithakar: Here , `preview_field.*` or `other_alias_field.*` doesn't have a column in a particular collection because it is `alias`. We have only one entry of alias fields in `directus_fields` collection. 

If the user passes 
`https://demo-api.directus.app/uploads/_/originals/{{trailer.filename}}` 
in URL template of `Preview` interface  then we can't get `trailer` field by using this solution.

So Let me know your thoughts.

rijkvanzanten: > Here , preview_field.* or other_alias_field.* doesn't have a column in a particular collection because it is alias.

That's not always true: a O2M is also an alias field, but that does in fact has support for `.*`

rijkvanzanten: Ideally the display component of an interface has the ability to let the parent know what data it relies on. 

I don't think fetching all data for all fields, just to make sure `preview` is able to show data is a good solution here. It's a giant performance overhead when the user will display the preview field.

We can also update the `display.vue` component of the preview interface to dynamically create and open the link once the user hovers / clicks on a button, instead of showing the preview inline.

rijkvanzanten:
@urvashithakar could you take a look at the points above? 🙂

urvashithakar:
Hello @rijkvanzanten,
I updated the PR with requested changes and the following flow. In items.vue[src/components] , when `alias` fields are present in the collection and if it has a dependency on other fields then it will fetch that fields using field.*

I hope it works for you.

rijkvanzanten:
@urvashithakar could you resolve the merge conflict? I don't know where it's supposed to go

bjgajjar:
Closed in favor of https://github.com/directus/app/pull/2152

PR #2033, opened 2019-09-05
Update german translation for 'on_create'

alperbalkaya:
Current translation is not fitting since it means 'on collection'


rijkvanzanten:
Thanks @alperbalkaya!

We manage the translations in [Crowdin](https://locales.directus.io), so I've updated it there 🙂

PR #2037, opened 2019-09-06
Fix #2035[Custom status was not able to filtered]

### LINKED ISSUE #2035: I can't see my custom status
I create a new status on mapping status but I can't see en table list
Screenshot_20190905_124530
Screenshot_20190905_124708

bjgajjar:
The default value of boolean was not set for `Status` interface. That cause the issue for `API`. 
This PR will emit the value on the created event.

### DIFF
--- a/src/components/form-inputs/toggle.vue
+++ b/src/components/form-inputs/toggle.vue
@@ -23,6 +23,10 @@ export default {
       default: null
     }
   },
+  created() {
+    // Emit input on created event to pass value to parent component
+    this.$emit("input", this.value);
+  },
   methods: {
     emitValue() {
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
This also means that the app always has some sort of change, which in turn always enables the save button even though there wasn't an actionable change. Isn't the API in charge of saving the default value if the app doesn't pass any?

bjgajjar:
> This also means that the app always has some sort of change, which in turn always enables the save button even though there wasn't an actionable change.

This change will not stage the value. Thus, the save button will not enable until any actionable change is not performed.

However, As per your recommendation, I am trying to add a condition on the filter of `API`.

PR #2043, opened 2019-09-09
Fixed typo in language files

steinarhovland:
Removed _is_ from 'Specific to translation interfaces, this o2m **is** stores multi-lingual content'


rijkvanzanten:
Thanks @steinarhovland!

@directus/team we have to make sure it doesn't get overwritten by the JSON files generated out by https://locales.directus.io. We might have to paste the file in there directly to prevent conflicts

PR #2059, opened 2019-09-13
wysiwyg image resize added

mustafa220:


vue.config.js
```diff
@@ -10,5 +11,13 @@ module.exports = {
     if (process.env.NODE_ENV === "development") {
       config.output.filename("[name].[hash].js").end();
     }
+  },
+  configureWebpack: {
```
rijkvanzanten: Is there any way we can make this quill plugin work without having to inject Quill in the global scope?

benhaynes: Are we still planning on moving to TinyMCE as our main/only WYSIWYG?

shealavington: Wasnt it a move to TipTap?

benhaynes: Haha, yeah... but more recently @rijkvanzanten has been thinking TinyMCE might be the more flexible.

rijkvanzanten: TipTap has way too much overhead unfortunately. TinyMCE is a little more plug and play and hands off for us

rijkvanzanten:
Closing this in favor of the new TinyMCE based wysiwyg editor

PR #2060, opened 2019-09-13
Fixes displaying users with avatars

### LINKED ISSUE #2058: User related fields are broken
As raised with the team, there's an issue displaying the user related data into table listing. Might be something related to avatar.

To reproduce the issue:

- Go to user profile
- Upload a profile picture
- Save
- Create a collection with created_by and modified_by and some fields
- Navigate to the collection
- Add a new entry
- Navigate to the collection again (at this point the table is already broken and can't be seen any data)

image

WoLfulus:
This fixes #2058

### DIFF
--- a/src/components/items.vue
+++ b/src/components/items.vue
@@ -404,7 +404,7 @@ export default {
 
         params.fields = params.fields.join(",");
       } else {
-        params.fields = "*.*";
+        params.fields = "*.*.*";
       }
 
       if (this.searchQuery) {


rijkvanzanten:
This might be a pretty big performance regression as it will now fetch 3 levels deep for all fields, not just the avatar field.

WoLfulus:
how do we proceed on fetching the avatar data then?

rijkvanzanten:
This would be the place to do it, but it needs to be a little smarter in when it fetches the extra level. This was changed from `*.*.*` to `*.*` on purpose because it made queries on larger tables waaaaaaay too slow

WoLfulus:
I'm unsure what to do in this case

benhaynes:
Aren't we planning on loading all directives data, including users, to the App data store? Wouldn't that mean we can just get the ID and the app can display the rest automatically?

bjgajjar:
@benhaynes  - Maybe https://github.com/directus/app/issues/2058#issuecomment-531164192 will help you :)

benhaynes:
Yeah, your `user` type.

I was actually thinking we load all `directus_users` once (on `app..load`) and write it to local storage. Then we don't have to load _any_ user data in all the other calls.

Weren't we thinking about doing this with most Directus system data, @rijkvanzanten ?

WoLfulus:
That's not an option I think. For example if all my "app" users are registered within directus system and I have 15.000 users?

It's probably better to just load on demand, store in app memory (and garbage collect when it's not in use), otherwise if I iterate over all pages I'll have 15.000 users fetched to memory

benhaynes:
I think I made the same argument to Rijk a while ago... and I feel like he had some crafty answer. We'll see what he says. 😄

rijkvanzanten:
That idea has been shelved since. @WoLfulus is right, it's not great for performance when you have a larger installation + you don't always have the proper permissions. You can have a user that doesn't have read access to any other user, which makes this useless.

benhaynes:
Gotcha. So is this good to merge in?

rijkvanzanten:
No, as per [my first comment](https://github.com/directus/app/pull/2060#issuecomment-531321423):

> This might be a pretty big performance regression as it will now fetch 3 levels deep for all fields, not just the avatar field.

rijkvanzanten:
Closed in favor of 61f12765

Issue #2050, opened 2019-09-11
Sticky menu


WoLfulus:
# Feature Request

Make the left bottom profile menu sticky.

## What problem does this feature solve?

I always keep hitting logout button or missing the action because menu expand/collapses on hover. 

## How do you think this should be implemented?

Make the menu expand on click "^" and collapse on click "V". It keeps its state

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

Issue #1236, opened 2019-01-09
🐰 Design Update — Community Feedback


benhaynes:
# 🐰 7.1 Design Update — Community Feedback
This ticket provides place to gather community feedback for our upcoming UX/UI update planned for v7.1. **Anyone is welcome (and encouraged) to post thoughts, comments, criticism, feedback, emoji, or ideas to this issue's thread.**

## What problem does this feature solve?

* Simplify the overall Application structure
* Abstract all (core and extension) pages into modules
* Allow for dedicated navigations per module
* Modularize the Info Bar content
* Apply consistent styling across the Application (eg: new styleguide)
    * Improve legibility and internationalization support (eg: RTL)
    * Improve theming by standardizing palette
    * Reassess component sizing for better responsiveness (eg: breadcrumb)
    * Simplify the form grid and standardize input styling
* Refactor of all keystone core components

## Would you be willing to work on this?
I already have! Below are the _initial_ design mockups for community review. These are works-in-progress and subject to change. 

> The primary architectural difference in the 7.1 UX is that left-sidebar has been split into a Module bar and a subordinate Navigation bar. This means that modules like File Library, User Directory, Activity, Settings, and any custom Module Extensions (previously "Pages") will have their own dedicated navigation.

![directus 7 1 browse](https://user-images.githubusercontent.com/522079/50868520-ed9f7180-137d-11e9-9484-e4a990f6a1fa.png)

![directus 7 1 detail](https://user-images.githubusercontent.com/522079/50868544-03ad3200-137e-11e9-861c-c421ef064e66.png)

![directus 7 1 detail info](https://user-images.githubusercontent.com/522079/50868566-1f183d00-137e-11e9-932c-05f63bf472cd.png)

![directus 7 1 browse files](https://user-images.githubusercontent.com/522079/50868552-10ca2100-137e-11e9-882f-7198d40aefb7.png)

![directus 7 1 dashboard](https://user-images.githubusercontent.com/522079/50868560-1a538900-137e-11e9-8ce9-83898b3b1858.png)

shealavington:
Will probably need a breadcrumb on the file navigation, this way if you head into multiple folders deep, you can easily come back up the tree without having to press 'back' 4 times, and visually track how deep you are into the file system quickly.

benhaynes:
[...]

PR #2063, opened 2019-09-15
Adds a button to quick navigate to collection settings

WoLfulus:
This is just a suggestion from something that really annoyed me when working all day long on collection setup. It's just displayed when currentUser has admin

This adds a small button on items to quickly navigate to collection settings instead of having to do big navigations to edit a collection fields:

- menu (please see #2050 too)
- admin settings 
- collections 
- collection

image

goes to 

image

### DIFF
--- a/src/routes/items.vue
+++ b/src/routes/items.vue
@@ -17,14 +17,6 @@
         >
           <v-icon :name="currentBookmark ? 'bookmark' : 'bookmark_border'" />
         </button>
-
-        <button
-          v-if="this.$store.state.currentUser.admin"
-          class="settings"
-          @click="editCollection()"
-        >
-          <v-icon :name="'settings_application'" />
-        </button>
       </template>
       <v-search-filter
         v-show="selection && selection.length === 0 && !emptyCollection"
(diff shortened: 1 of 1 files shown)


benhaynes:
Again, I love the idea... But this should be solved soon with this update:

https://github.com/directus/app/issues/1236

rijkvanzanten:
Will it @benhaynes? I don't see this button in those screenshots? Or am I missing something?

I'd like to merge this in regardless, as it's a really useful feature to have.

benhaynes:
Yeah, it's in the updated design files... which have changed quite a bit since posting them here. "Settings" will be a conditional module for admins, so it'll always be available directly.

benhaynes:
We can merge this in until the new design is released...

WoLfulus:
I thought about moving to a config button on the collection listing itself (left side panel) so you'll be able to access both the collection items and the collection settings from anywhere.

Thoughts?

benhaynes:
Having this in the Collections list is a bit more complicated since that can have a `nav_override` value. Again, I think the _right_ solution is the new Module Bar design... but we can merge this in in the meantime. :)

PR #2064, opened 2019-09-15
Allows sticky the user menu

WoLfulus:
Solves #2050

I couldn't find a pin icon though

https://i.imgur.com/XGlQwGF.gifv

### DIFF
--- a/src/components/sidebars/nav-sidebar/user-menu.vue
+++ b/src/components/sidebars/nav-sidebar/user-menu.vue
@@ -1,9 +1,11 @@
 <template>
-  <div class="user-menu">
+  <div :class="menuClass">
     <header>
       <v-avatar :src="avatarURL" :alt="fullName" :indicator="true" class="avatar" />
       <span class="no-wrap">{{ fullName }}</span>
-      <v-icon name="expand_more" />
+      <button @mouseover="overLock = true" @mouseleave="overLock = false" @click="locked = !locked">
+        <v-icon :name="lockIcon" />
+      </button>
     </header>
     <div class="links">
       <nav-menu :links="userLinks" class="menu" />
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @WoLfulus !! This actually won't be an issue in the new design (coming out in about a month) since all of these user menu items will get moved to the module menu, which is always visible.

Ref:

benhaynes:
I like the effect, but I don't know about the icon since we're using that elsewhere for auth. Rijk?

rijkvanzanten:
I agree on the icon. I also feel like this UX problem might have a "nicer" solve than a user interaction. EG only hide the menu if the cursor is 100px out of the menu or after 2 seconds or something

WoLfulus:
This isn't the problem though, I kept clicking on the logout button (because I'm used to "click to expand") and because it expands too quickly, the logout button is pressed instead of the header. Another option was to only expand on header click and whenever you click the header again it collapses (persisting across pages).

This gets really annoying after hitting logout 10x or so

rijkvanzanten:
Ohhhhh! I see!

I wouldn't mind making it a click to open to be honest. @benhaynes ?

benhaynes:
The hover has never bothered me, but if you both agree that `onclick` is better we can update it. We just need to be careful about too many changes to this UX since potentially tens of thousands of other users will have to also get used to a new behavior.

For the icon, we can just use the same/current arrow, but rotate it 180° in the "open" position.

WoLfulus:
If we change to onclick we can just keep 
image

rijkvanzanten:
We can do a `◀︎` for closed, and `▼` for open

rijkvanzanten:
User menu has since been replaced with a new sidebar:

image

PR #2089, opened 2019-09-24
Webhook Directory

bjgajjar:
Ref. of https://github.com/directus/api/issues/29


benhaynes:
Binal and I discussed adding a `status` field to this so it's easier to turn on/off.

PR #2093, opened 2019-09-25
Fix directus/api#1211 [Use the dynamic * from settings table for fetching the relational data]

bjgajjar:
Fixed https://github.com/directus/api/issues/1211


bjgajjar:
Closing this PR - as per the discussion with @rijkvanzanten it's easy to be a break. If we change it to be 1 level deeper, the queries will be an extremely slow and most likely timeout. If we set it to 1 level fewer interfaces/layouts start breaking

PR #2109, opened 2019-10-02
Fix #2088 [O2M > Translation Interface : Removed diff]

### LINKED ISSUE #2088: id lost on second level translations on updates
# Bug Report

This happens with APP 7.10.0 and API 2.5.0 on https://directus.app.

When a child collection has translated content, changing existing translated content is missing the id and language_code from the payload as sent from the app, which results in the API saving a null language code, which in turn breaks the content.
The translated content section can then not be viewed anymore and instead of the translated content the error: "warning There are no languages in the languages collection"

## Steps to Reproduce

Create collections demo_languages, demo_articles_sections_translations, demo_articles_sections and demo_articles (some fields removed for simplicity)
I also set up language_code as a text as well as a "language" interface and had the same outcome.

```
All fields are mandatory

demo_languages
      id
      language_code (text)
      name (text)

and add entries for EN/English FR/French

demo_articles_sections_translations (hidden collection, only access from demo_articles)
      id
      parent_id (m2o, relation to demo_articles_sections, hidden)
      language_code (text, hidden)
      text (text)

demo_articles_sections (hidden collection, only access from demo_articles)
      id
      parent_id  (relation to demo_articles, hidden) 
      label (text)
      translation   (translation, languages collection = demo_languages, languages PK = language_code, Languages Name Field = name, Translation Language Field = language_code, default language = EN)

demo_articles
      id
      label
      sections (one to many)

```
creating a new demo_articles item with a new demo_articles_sections and new translated text works and the payload looks like this:
````
{"label":"parent label 2","sections":[{"label":"section label 2.1","translation":[{"language_code":"en","text":"english text 1"}]}]}
````

but when I attempt to update the English text from the previously created demo_articles to something else, the payload looks like:
```
{"sections":[{"id":2,"translation":{"0":{"text":"english text 1 updated"}}}]}
```
and is missing the id and language code and creates new entries in demo_articles_sections_translations with a null language code.

When this is then received by the API, it stores invalid data and the next time the item is opened it shows "There are no languages in the languages collection" instead of the translations.

## Expected Behavior

When updating demo_articles_sections_translations from with a demo_article, the translations are updated.

I also believe that the language code, which is marked as required, should fail with a null value, instead of storing new entries with null language_code values in demo_articles_sections_translations

## Actual Behavior

a new entry with a null language code is saved with the correct parent_id and this means the interface cannot display the translations anymore and the data cannot be fixed from withing the parent collection, either.

I also reproduced it against the latest develop to check the fix for https://github.com/directus/api/issues/1227 

When I browse directly to the hidden collection/item and update eg the English translation  from there, it works fine and the payload from the APP is :
``` {"content_translations":[
{"id":2,"language_code":"en-US","header":"carrot soup","text":"Updated to Carrot soup recipe","created_by":{"userdetails":"..."}]},"created_on":"2019-10-01 01:08:41","modified_by":{"userdetails":"..."}]},"modified_on":"2019-10-01 01:08:41"},
{"id":3,"language_code":"fr-FR","header":"soupe aux carottes","text":"recette de soupe aux carottes","created_by":{"userdetails":"..."}]},"created_on":"2019-10-01 01:08:41","modified_by":{"userdetails":"..."}]},"modified_on":"2019-10-01 01:08:41"}]} ```

itsmerhp:
Skipped diff for translate interface inside the O2M interface.

Fixes #2088

### DIFF
--- a/src/interfaces/one-to-many/input.vue
+++ b/src/interfaces/one-to-many/input.vue
@@ -455,6 +455,8 @@ export default {
 
               if (type === "json") {
                 delta[key] = after[key];
+              } else if (type.toLowerCase() === "translation") {
+                delta[key] = after[key];
               }
             });
 
@@ -507,7 +509,6 @@ export default {
           $delete: true
         };
       });
-
       this.$emit("input", [...newValue, ...deletedRows]);
     }
   }
--- a/src/interfaces/one-to-many/input.vue
+++ b/src/interfaces/one-to-many/input.vue
@@ -455,7 +455,7 @@ export default {
 
               if (type === "json") {
                 delta[key] = after[key];
-              } else if (type.toLowerCase() === "translation") {
+              } else if (type === "translation") {
                 delta[key] = after[key];
               }
             });


rijkvanzanten:
Ref #2088

PR #2110, opened 2019-10-02
pass through all translations data on child collection

markus-mnm:
I am still testing this, and it might not work in all situations and preferably I would only add the primary key of all the updated translations to the diff of the translation instead of adding the whole payload. But I am not sure if this is possible and worth it. (I haven't checked yet)

### DIFF
--- a/src/interfaces/one-to-many/input.vue
+++ b/src/interfaces/one-to-many/input.vue
@@ -453,7 +453,7 @@ export default {
 
               const type = fieldInfo.type.toLowerCase();
 
-              if (type === "json") {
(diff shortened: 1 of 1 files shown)


markus-mnm:
this is a potential fix for https://github.com/directus/app/issues/2088

rijkvanzanten:
This might be a duplicate of https://github.com/directus/app/pull/2109/files

markus-mnm:
Closing this PR because the other PR  https://github.com/directus/app/pull/2109/files also resolves this

Issue #2106, opened 2019-09-30
Checkboxes don't appear in "Add New" modal popup


rdsubhas:
### Given

* Collection `refunds`
  * field `refund_methods`, type `Checkboxes`, values: `{ "option1": "value1", "option2": "value2", ... }`
  * field `company` – M2O with companies
* Collection `companies`
  * field `refund_policies`, type `O2M` with `refunds`, Allow creation

### Problem

If I normally go to the `refunds` page – checkboxes are clearly visible. This is fine:

![image](https://user-images.githubusercontent.com/1570745/65907852-6be8eb80-e3c5-11e9-91dd-52a5701ef834.png)

But if I go to the companies page and try to `Add New`, checkboxes get hidden:

![image](https://user-images.githubusercontent.com/1570745/65907785-4f4cb380-e3c5-11e9-91cd-8a741ae002a9.png)

![image](https://user-images.githubusercontent.com/1570745/65907737-33491200-e3c5-11e9-9c8e-0c64b19d84c4.png)

Changing that single field from `Checkboxes` to `Multi-select` shows up all the options.

<img width="1421" alt="image" src="https://user-images.githubusercontent.com/1570745/65907686-1a406100-e3c5-11e9-9f76-96c56011f24d.png">

But for some reason, `Checkboxes` values alone get hidden somehow in the modal. It's not a CSS problem or anything. Inspecting shows an empty div.

<img width="1417" alt="image" src="https://user-images.githubusercontent.com/1570745/65907986-b702fe80-e3c5-11e9-8cb5-d506c63ff63c.png">

### Directus version

Running latest versions.

* Api `v2.6.0`
* App `v7.11.0`

bjgajjar:
Fixed in https://github.com/directus/app/pull/2111

rdsubhas:
Thanks @bjgajjar !

PR #2111, opened 2019-10-04
#2106 : Undefined error solved in initChoices

itsmerhp:
In this PR below two issues resolved :
- O2M > Checboxes interface > Create: Undefined error for initial choices
- O2M > Checboxes interface > Update: Checkboxes choices not updating properly

### DIFF
--- a/src/interfaces/one-to-many/input.vue
+++ b/src/interfaces/one-to-many/input.vue
@@ -453,7 +453,7 @@ export default {
 
               const type = fieldInfo.type.toLowerCase();
 
-              if (type === "json") {
+              if (type === "json" || type === "transition" || type === "array") {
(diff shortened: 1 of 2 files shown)


src/interfaces/one-to-many/input.vue
```diff
@@ -453,7 +453,7 @@ export default {
 
               const type = fieldInfo.type.toLowerCase();
 
-              if (type === "json") {
+              if (type === "json" || type === "transition" || type === "array") {
```
rijkvanzanten: There is no `transition` type. Is this supposed to be `translation` (just like that PR you made)?

itsmerhp: @rijkvanzanten Change implemented!

PR #2113, opened 2019-10-05
Adds support for default locale on first load

WoLfulus:


src/app.vue
```diff
@@ -63,6 +64,9 @@ export default {
     VNavSidebar,
     VNotification
   },
+  data: () => ({
+    loaded: false
+  }),
```
rijkvanzanten: Lets use this structure:

```js
data() {
  return {
    loaded: false
  }
}
```

so it's consistent with the rest of the app

Also, lets rename `loaded` to `languageLoaded`

WoLfulus:
@rijkvanzanten I see you already changed what you requested, but it's still marked as changes requested. There's more stuff to do in this one?

rijkvanzanten:
I don't think so! I didn't have time to do a final test yesterday so had to leave it for a bit

PR #2141, opened 2019-10-16
Add "Download" option for file interface

PostPollux:
I added a "Download" option for the file interface in the options menu (three small dots).

We needed this for a project where some maintainers are supposed to upload and maintain excel files that contain translations for an app.
Now if they want to quickly change one translation it's easier to just quickly download, change and reupload it again, than first searching for that file on the harddrive...
Of course it was possible to download the file by going to the files directory. But I think having that shortcut is much more convenient.

Link for testing with gitpod: http://gitpod.io#https://github.com/PostPollux/app/tree/add_file_download_button

Here is what it actually looks like:

download_option


PostPollux:
> I like this a lot more in that contextual menu, instead of a button. Looks good to me!

Yes, me too. The button was to big. And I noticed that it jumped into a second row with the small layout. The context menu is a much better fit...

src/interfaces/file/input.vue
```diff
@@ -209,6 +214,9 @@ export default {
 
       this.newFile = false;
     },
+    downloadFile() {
+       window.open(this.image.data.full_url)
```
bjgajjar: IMO, `Download` functionality must download the file instead of opening in the new tab.
For that, we can use the `download` attribute of `a` tag. 

Although, I know the `Download` is a `button`; but we can prepend `a` tag at a run time to this `button` tag.

PostPollux: Thanks for the feedback! I agree. That was actually what I wanted to achieve in the first place, but I couldn't get it to work, so I used that workaround with window.open()

Unfortunately I'm not very versed in web development. I'm a complete beginner in vue, java script and php.

So this is what I already had tried:
 
This one just adds an ugly  "download" link. When I click it the exact same is happening. If it can display the file format it will display it. If not the download dialog will open. The only difference is that it doesn't open in a new window. AND of course that when I right click on the link and hit "save link as..." it will actually save the file and not some html file:
```
<a v-if="value" :href="image.data.full_url" download> download </a>
```

This was a dedicated button instead of a option in the context menu. I used an "a"-tag with the "download" keyword to enclose the button. Unfortunately klicking on that button did NOTHING. But right klicking anywhere on the button and choose "save link as.." worked as expected.
```
<a v-if="value" :href="image.data.full_url" download>
        <v-button  type="button" :disabled="readonly">
        <v-icon name="file_download" />
        {{ $t("file_download") }}
        </v-button>
</a>
```

I also tried window.location instead of window.open(). It opens the link in the same window and not a new one. But it behaves the same. As soon as the Browser can open that file, it will display it.

```
downloadFile(){
        window.location = this.image.data.full_url;
    }
```

I had no idea how to fix this. And when I was thinking about it, I remebered that I have seen that pattern so often on the web where the website is telling you, that you have to right click on the following link and choose "save link as..." to download the file. 
So I was wondering that it's maybe not possible.

I can easily change that it opens in the same browser tab. But for always invoking a download dialog I need help.
Also I don't know how to wrap an  "a"-tag with the "download" keyword around that option in the context menu.

So any help is very appreciated.

rijkvanzanten:
I agree that having a native `<a>` tag is preferable in this case, but seeing the limitations of the current implementation of the context menu, this is the best we can do. I'm currently refactoring a lot of the core components; when I get to contextual menu's I can revisit this and provide a proper alternative 🙂 

Thanks @PostPollux! And apologies for the delay in merging this 🙂

yagobski:
This change is not merged with https://github.com/directus/directus/tree/master/public

rijkvanzanten:
It was merged into the app on master so it'll be included in the next release

PR #2153, opened 2019-10-23
Fix #2119 [Fields can not be unchecked in the Permissions "On Creation" ]

### LINKED ISSUE #2119: Fields currently can not be hidden
Fields can not be unchecked in the Permissions "On Creation"
The expected behaviour should be that they are hidden when unchecked, but use default Value when created.

Screenshot_20191010_050110

Fields unchecked in the other groups(like Published etc) seem to only hide the Value, but not the field itself in the App.
The expected behaviour should be that they are hidden. And a required field needs to use the original Value so the form can still be sent.

**Update:**
After talking to Binal i made these 2 Videos trying to show what i am talking about:
https://youtu.be/KkrbX4fBUvM
https://youtu.be/5ej04il8AMs

My workaround for the time being was to remove this field and have a second Table Validate the publication status of my Blog entries. Not a direct fix i kinda thought i was using the Backend in someway wrong.

urvashithakar:


rijkvanzanten:
This was fixed in cf1d22a

PR #2161, opened 2019-10-25
Fix #2156[ Nav override not showing titles ]

### LINKED ISSUE #2156: Nav override not showing titles
# Bug Report

## Steps to Reproduce

1. Go to roles > role > nav override
2. Create a section with title and custom links
3. See the left sidebar
4. The title is not showing up

## Expected Behavior

The sections defined in nav override have a title

## Actual Behavior

The sections do not have a title

image
image

## Technical Details

- Device: [eg: Desktop, iPhone6, etc]
- OS: [eg: MacOS 10.13.6, Windows 10.1803]
- Web Server: [eg: Apache 2.4.37]
- PHP Version: [eg: 7.2.0]
- Database: [eg: MySQL 8.0.12]
- Install Method: [eg: cloned `master` branch]

urvashithakar:
Fixes #2156

### DIFF
--- a/src/components/sidebars/nav-sidebar/nav-menu.vue
+++ b/src/components/sidebars/nav-sidebar/nav-menu.vue
@@ -1,5 +1,6 @@
 <template>
   <div v-if="links.length > 0" class="nav-menu">
+    <h3 v-if="title" class="style-4">{{ title }}</h3>
     <nav>
       <ul>
         <li v-for="{ path, name, target, icon, color } in links" :key="path">


rijkvanzanten:
I'm pretty sure the titles in the nav sidebar where removed on purpose in the design update. @benhaynes?

lluishi93:
@rijkvanzanten Hi! I was the one who opened this issue. We're using directus with a client that wants a lot of components. Allowing to separate them by sections, and highlighting their types with a title it's a necessary thing, specially when less technically-oriented people starts working with the cms.

But, if the title was removed, then the documentation should be updated to reflect that: https://docs.directus.io/guides/roles.html#creating-roles

I think that most people would agree that the title won't hurt, and if you just don't put the title in the json, then it should not appear, but I don't think that removing that option completely is a good idea (IMO of course)

benhaynes:
I agree it's a useful feature/option — and we'll merge this back in. It's just the _system_  sidebars that won't have any group titles (just HRs to delineate)... so only nav_override will have this option. 

@urvashithakar — can we confirm that this section header collapses (is gone) when the `title` string is empty? After that we can merge and I will polish up the styling to match this:

Screen Shot 2019-10-26 at 9 41 02 AM

PR #2173, opened 2019-10-29
Wysiwyg

rijkvanzanten:


dario30186:
Hi @rijkvanzanten ,
I just install last docker images for app and api and the api still uses Quilljs.
When TinyMce will ba available in docker image? Are you going to release a new api docker image soon? If not, my team will try to add a custom interface. 
We already use Tinymce with directus 6 for another project ( it was the default html editor before).

Thank you

rijkvanzanten:
It'll be in the next release, can't promise any dates on that

PR #2182, opened 2019-11-04
Now ZEIT deployment config

Autr:
Hi, thought this might be useful for some people - a config json for resolving routes when deploying to ZEIT now

### DIFF
--- a/now.json
+++ b/now.json
@@ -1,0 +1,34 @@
+{
+  "name": "directus-app",
+  "alias": [],
+  "builds": [
+    {
+      "src": "/dist/**",
+      "use": "@now/static"
+    }
+  ],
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Thanks @Autr! As of Directus v8 and up, the app can't be used standalone, so deploying just the application (without the API), won't do anything. Thanks anyway!

Autr:
@rijkvanzanten do you mean the frontend is no longer going to be standalone? I am currently running my frontend via ZEIT, and the API via a LAMP server - will this no longer be possible?

rijkvanzanten:
That's correct. The admin app is always served from the API on the LAMP server moving forward

PR #2199, opened 2019-11-12
Make the using of the telemetry service configurable

arboehme:
The new config option for the APP to enable / disable the using of telemetry service  (see https://github.com/directus/docs/issues/167)


rijkvanzanten:
The config file for the application has been removed in v8. I've implemented the same functionality in 4ccfb981.

PR #2203, opened 2019-11-13
compact translations for the menu

vis97c:
I've been translating the project to spanish but recently i notice that the translations for the menu were to long they weren't wrong at all just too long for the side menu, this is my new proposal, thanks in advantage.

This proposal keeps the same sense whitin the app context
Proposal


rijkvanzanten:
Thanks!

PR #2205, opened 2019-11-13
Feature: Copy raw collection name

grandsilence:
Useful for development when collections are localized.
copy-collection-name


benhaynes:
Thank you!! A few points:

* If we were to add this, we would probably just want to add the tooltip directly to the collection name (no need to add visual clutter)
* We're trying to keep the non-admin side as clean/simple as possible... and this seems like more of an admin feature. 
* As a developer, if you are on any of the collection pages (browse or detail), can't you simply look in the browser's URL bar to see the non-localized database collection key?

grandsilence:
You are right. May be I'll make the same pull request but on icon of collection (tooltip + copy on click). Anyways I can use URL.

Issue #422, opened 2018-05-22
Implement translations for collection and field names


rijkvanzanten:
When fields are used, a generated `name` key should be added that uses the translation for the current locale, or fallback to the `formatTitle()` helper if a translation key doesn't exist.

benhaynes:
@rijkvanzanten don't we have this with `directus_fields.translation`?

rijkvanzanten:
We have the API infrastructure in place, but it isn't used in the app yet.

benhaynes:
@rijkvanzanten ?

rijkvanzanten:
no news

rijkvanzanten:
Thats exactly the idea @binary-koan! We want to have a translation interface that will save to this particular json structure.  Once we have that, we can "open up" this field and start using the translations everywhere 😄

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

PR #2207, opened 2019-11-13
Feature: Global field localization using [lang-LANG]/index.json

grandsilence:
# Global field localization

In `[lang-LANG]/index.json`
1. Collection field name: `"field-[field_name]": "name",` (preferred)
2. Global field name for all collections: `"field-[collection]-[field_name]": "name",`

New helper method for it:
1. `$helpers.formatField(fieldAsObject)` where `field` is `object` with `field.field` and `field.collection`.
2. `$helpers.formatField(fieldName, collectionName)` where `fieldName` is `string` and `collectionName` is `string`.


benhaynes:
Thank you @grandsilence!! We'll take a look asap!

rijkvanzanten:
I'll try to get this in before the final release of V8.0.0 🚀

everyx:
sorry, is this related #422 ?

grandsilence:
Yeah, it’s implementation of #422

everyx:
So we can set collection meta translate like this? https://github.com/directus/app/issues/422#issuecomment-455798477

image

rijkvanzanten:
@everyx We might use the repeater interface here since we have that now, it's a little more user friendly than requiring raw json

src/helpers/format-field.js
```diff
@@ -0,0 +1,43 @@
...
+    (fieldIsString && !field.length)
+  ) {
+    console.error(`fieldLocalize: Invalid value of field: ${JSON.stringify(field)}`);
+    return "???";
+  }
+
+  const fieldName = fieldIsObject ? field.field : field;
+  const fallback = Vue.$helpers.formatTitle(fieldName);
+
+  if (!Vue.$te || !Vue.$t) {
```
rijkvanzanten: We're using this helper in the Directus app which will always have the Vue-i18n plugin available. This check is redundant

src/helpers/format-field.js
```diff
@@ -0,0 +1,43 @@
...
+
+  if (collectionName) {
+    const collectionField = `fields-${collectionName}-${fieldName}`;
+    if (Vue.$te(collectionField)) {
+      return Vue.$t(collectionField);
+    }
+  }
+
+  // Global field name (all collections)
+  const globalField = `fields-${fieldName}`;
```
rijkvanzanten: Is this how they're scoped in the app right now? 🤔 Maybe we should scope them as `collection.field` at all times. I can't think when we'd have a field name that should be global

src/helpers/format-field.js
```diff
@@ -0,0 +1,43 @@
+import Vue from "../main";
+
+export default function formatField(field, collection) {
```
rijkvanzanten: I'm wondering if we should make this more strict in only allowing field and collection to strings, and make them both required 🤔

PR #2220, opened 2019-11-21
Wysiwyg use file picker/uploader instead of selector for image imports

jacobrienstra:
Right now we use a file selector, so instead I hooked up the file input to the WYSIWYG file field so you can upload or select an existing image. 
Feel free to ask for any changes or if you have any feedback!


rijkvanzanten:
Tweaked a tiny thing:

It didn't empty out `selectedFile` when hitting done, which meant that the upload component was already "done uploading" when you added an additional file. I fixed that by setting `selectedFile` to `null` after the callback to TinyMCE is done

PR #2236, opened 2019-11-25
Do raise an obscure error if APP_URL is undefined when starting a dev server

samvasko:
If the APP_URL env variable was not specified the web server would fail with a confusing message.

    TypeError: Cannot read property 'upgrade' of undefined

This MR makes the said environmental variable optional.

We should also add something to docs about the URL. https://docs.directus.io/advanced/source.html#application-source


rijkvanzanten:
I don't know if making the variable optional would solve anything though.. That would mean the app starts up fine, but you won't be able to do anything as the app itself will start fetching data from a non-existing proxied API 🤔  It'd mean that we'd delay the weird error to a later point

samvasko:
Yeah, I just took a crack at it. If you think this is the only way to use it. 

I will change this so it asserts that the environmental variable is there.

samvasko:
I was hoping to redirect to directus demo server but I am because its using 301 requests I get a error on authenticate post request.

rijkvanzanten:
It should work when we use HTTPS instead of http 😄 the 301 is the redirect from http to https

PR #2326, opened 2019-12-11
Add field, interface, and layout translations to Crowdin

rijkvanzanten:
This PR allows Directus system fields, interfaces (and options), and layouts to be translated through Crowdin.


rijkvanzanten:
@everyx 😄 🎉

Issue #1885, opened 2019-07-25
Add "Other" option to Dropdown Interfae


bjgajjar:
We need an interface option "Other" for Dropdown that allows the user to type another value into a text input.

benhaynes:
While we usually close Feature Requests — let's leave this open for a bit for visibility. We'll need to add this option in (shouldn't take more than 30m or so) so the API password feature is complete.

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

rijkvanzanten:
The `v-select` core component that’s used by the dropdown interface already has an other prop that allows you to provide your own value. Adding support for allow-other in the interface would mean:
• Adding the option to the meta.json file of the dropdown interface
• Passing the value of that option to the v-select in the `input.vue` file of the dropdown interface ( `:other="options.allowOther"` )

ozcanoguz:
Hello  @rijkvanzanten ,

I've tried to do it with your instructions but i think your comment is invalid here. Tried to achieve this via : 

Adapted "src\interfaces\dropdown\input.vue"
```
<v-select
    v-else
    :id="name"
    :value="value"
    :disabled="readonly"
    :options="choices"
    :placeholder="options.placeholder"
    :icon="options.icon"
    :other="options.allowOther"
    @input="$emit('input', $event)"
  ></v-select>
```
and
"src\interfaces\dropdown\meta.json"
```
"allow_other": {
      "name": "$t:allow_other",
      "comment": "$t:allow_other_comment",
      "interface": "toggle",
      "default": false
    }
```

Add options option has been shown in collection creation, but widget itself does not allow to add. Additionally i've found this SO entry that mentions v-select does not allow that option :
https://stackoverflow.com/questions/49323169/input-custom-text-into-v-select
Vue documentation offers to use "taggable" "multiple" 
https://vue-select.org/guide/values.html#tagging

rijkvanzanten:
The problem here is that you use `options.allowOther` in `input.vue`, but `allow_other` in the meta.json 🙂 

Either rename `allow_other` in meta.json to `allowOther`, or use `options.allow_other` in `input.vue`.

ozcanoguz:
Hello @rijkvanzanten , thanks for feedback! Bad typo :)

[...]

PR #2363, opened 2019-12-25
Fix app #1885[Dropdown add other options]

ozcanoguz:
#1885 functionality has been added by this piece of code.
Creation of field/interface dropdown
creation
Available Values List - Using Free Field
available_values
Listing the item
listed_dropdown_with_freetext

### DIFF
--- a/src/interfaces/dropdown/meta.json
+++ b/src/interfaces/dropdown/meta.json
@@ -7,6 +7,12 @@
     "length": 100
   },
   "options": {
+    "allow_other": {
+      "name": "$t:allow_other",
+      "comment": "$t:allow_other_comment",
+      "interface": "toggle",
+      "default": false
+    },
     "choices": {
       "name": "$t:choices",
       "comment": "$t:choices_comment",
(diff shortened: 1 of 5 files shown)


rijkvanzanten:
@benhaynes do we consider 'allow other' an advanced option? I'm wondering what position would be best for it in the interface options pane. I think `choices` should definitely be first, followed by placeholder. Allow other could be moved under `advanced`, but I'm not sure if it's _that_ advanced 🤔

benhaynes:
No, `allow_other` seems pretty standard to me. No need to hide it under advanced.

src/interfaces/dropdown/meta.json
```diff
@@ -7,6 +7,12 @@
     "length": 100
   },
   "options": {
```
rijkvanzanten: Lets make the order of these 

```
choices
placeholder
allow_other
icon
formatting
```

ozcanoguz:
Change has been done. FYI @rijkvanzanten

PR #2368, opened 2019-12-28
disable slug generate when editing, fix #2335

### LINKED ISSUE #2335: Slug interface should not regenerate slug if it already generated when mirrored field changes
Hi,

Slug interface is usually used to generate url and mirrored `title` field for a website, if a title is changed, the site url should keep same as before the first generated, so can we disable regenerating for slug interface when mirrored field changes, or add a configuration option.

everyx:
When item is created, we should keep the slug don't auto generate mirror field 

Fixed #2335


rijkvanzanten:
We should base this on an interface option. You might want to be able to edit a slug after the fact.

PR #2369, opened 2019-12-29
User browser language as default if none is set

### LINKED ISSUE #2087: Initial login page language form window.navigator.language
Hi, guys,

I am from China, and I can't find anyway to change the login page language, so can we use the default language from `window.navigator.language`? it's more flexible and friendly. :)

everyx:
When we can't set the loocle according to the API, use a user preferred language before page rendered from `window.navigator.language` is more friendly for users.

Fixes #2087


everyx:
I'll do this when I have time some days later.

everyx:
bug

rijkvanzanten:
Hi @everyx, I updated the branch 🙂

PR #2381, opened 2020-01-03
Create language ES-419 Spanish (Latin America)

andgar2010:
Hello,
there is the spanish (Latin America) translation of the files.
- index.json
- layouts.json
- interfaces.json


rijkvanzanten:
We're using Crowding for managing translations, and are thus bound to the languages supported in there. Unfortunately there is no Latin America Spanish by default in Crowdin. Does any of these languages come close or should we consider adding a custom language @andgar2010?

https://support.crowdin.com/api/language-codes/

andgar2010:
> We're using Crowding for managing translations, and are thus bound to the languages supported in there. Unfortunately there is no Latin America Spanish by default in Crowdin. Does any of these languages come close or should we consider adding a custom language @andgar2010?
> 
> https://support.crowdin.com/api/language-codes/

Can you add Latin American language, without Crowdin dependence?

rijkvanzanten:
I should be able to add a custom one, I was just wondering if there was a more specific dialect this applied to 🙂

PR #2386, opened 2020-01-06
Feature: Add Tree-Layout 🌲

Nitwel:
Adds support for tree like database structures.

## Features:
- displays items relative to each other.
- changeable icons depending on the usecase. (e.g. groups of people or files in folders)
- support for merging additional related content.
- adds trees to directus to make it greener and more environment friendly.

🌲🌳🌲🌳🌲🌳🌲🌳 

## Preview:
grafik

### DIFF
--- a/src/layouts/tree/Leaf.vue
+++ b/src/layouts/tree/Leaf.vue
@@ -1,0 +1,86 @@
+<template>
+  <div class="leaf">
+    <div v-show="!last" class="line"></div>
+    <div class="dot"></div>
+    <v-icon v-if="icon" class="icon" :name="icon"></v-icon>
+    <div class="title" @click.stop="parent.openItem(item.id, true)">{{ title }}</div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "Leaf",
+  props: {
+    item: {
+      type: Object,
+      default: null
+    },
+    parent: {
+      type: Object,
+      default: null
+    },
+    last: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data() {
+    return {};
+  },
+  computed: {
+    title() {
+      return this.item[this.parent.viewOptions.friendsTitle];
+    },
+    icon() {
+      return this.parent.viewOptions.friendIcon;
+    }
+  },
+  created() {},
+  methods: {}
+};
+</script>
+
+<style lang="scss" scoped>
+.leaf {
+  position: relative;
+  display: flex;
+  align-items: center;
+
+  .dot {
+    z-index: 1;
+    position: absolute;
+    width: 10px;
+    height: 10px;
+    border: 2px solid var(--darkest-gray);
+    background-color: var(--white);
+    border-radius: 50%;
+    float: left;
+    top: 15px;
+    left: -10px;
+    transform: translate(-50%, -50%);
+  }
+
+  .line {
+    position: absolute;
+    width: 2px;
+    height: 100%;
+    background-color: var(--dark-gray);
+    float: left;
+    top: 15px;
+    left: -11px;
+  }
+
+  .icon {
+    margin-right: 5px;
+  }
+
+  .title {
+    cursor: pointer;
+    height: 30px;
+    font-size: 16px;
+
+    display: flex;
+    align-items: center;
+  }
+}
+</style>
--- a/src/layouts/tree/Leaf.vue
+++ b/src/layouts/tree/Leaf.vue
@@ -1,7 +1,5 @@
 <template>
   <div class="leaf">
-    <div v-show="!last" class="line"></div>
-    <div class="dot"></div>
     <v-icon v-if="icon" class="icon" :name="icon"></v-icon>
     <router-link class="title" :to="root.createLink(item.id, item.__collection__)">
       {{ title }}
@@ -49,32 +47,6 @@ export default {
   position: relative;
   display: flex;
   align-items: center;
-
-  .dot {
-    display: none;
-    z-index: 1;
-    position: absolute;
-    width: 10px;
-    height: 10px;
-    border: 2px solid var(--darkest-gray);
-    background-color: var(--white);
-    border-radius: 50%;
-    float: left;
-    top: 15px;
-    left: -10px;
-    transform: translate(-50%, -50%);
-  }
-
-  .line {
-    display: none;
-    position: absolute;
-    width: 2px;
-    height: 100%;
-    background-color: var(--dark-gray);
-    float: left;
-    top: 15px;
-    left: -11px;
-  }
 
   .icon {
     margin-right: 5px;
--- a/src/layouts/tree/Leaf.vue
+++ b/src/layouts/tree/Leaf.vue
@@ -26,9 +26,6 @@ export default {
       default: false
     }
   },
-  data() {
-    return {};
-  },
   computed: {
     title() {
       var connections = this.root.viewOptions.connections;
@@ -43,9 +40,7 @@ export default {
       if (!collection.icon) return this.$store.state.collections[this.item.__collection__].icon;
       return collection.icon;
     }
-  },
-  created() {},
-  methods: {}
+  }
 };
 </script>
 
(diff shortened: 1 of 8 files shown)


rijkvanzanten:
Hey! This is awesome! 😄 

Like we discussed just now on Meet, lets:

* Add support for drag and drop
* Add a "Template" option so you can configure what is shown for the value
* Sort everything alphabetically based on the output of the template
* Use collection icon instead of custom configurable icons for everything
* Use drop down arrows to indicate existence of sub-items

image

(Note how the nested item has a different icon because it's from a different collection

* Merge parent and "friends" into single repeater
* Pre-fill "friends" with all available m2o's to the current collection (based on relations)
* Every relationship needs a display template
* Display template should be pre-configured to be first non-primary key field
* If repeater doesn't fit well inside the sidebar (likely), consider adding a "edit relations" button that opens a modal with the repeater in it.
* Rename `friends` to `relations`

rijkvanzanten:
Alsoooooo.... can we pleaaaaaaaaseeeeeeeeeee.......


image

benhaynes:
Ahh, using the collection icon... genius! 😄 

I think this layout is a good time to implement the "Template String" in the collection Settings. We can use that in a lot of places, and it will be too hard to add them as layout options for any possible relationships. If the collection doesn't have one set, we can fall-back to the first (non system) field.

src/layouts/tree/layout.vue
```diff
@@ -0,0 +1,144 @@
...
+  destroyed() {
+    //document.removeEventListener("scroll", this.scroll);
+  },
+  methods: {
+    scroll(event) {
+      var timeline = this.$refs.tree;
+      var toBottom = timeline.offsetTop + timeline.clientHeight - window.innerHeight - event.pageY;
+
+      if (toBottom < 100 && !this.lazyLoading) {
+        this.$emit("next-page");
```
rijkvanzanten: Seeing we don't load relational items up front, and we only fetch like 1 field of data, can't we fetch everything on load? If the collection is ~1000+ items, a tree view doesn't necessarily make sense to use to begin with. (@benhaynes we could say that collections > `x` items are restricted to a given layout. Viewing a million records in a layout like this doesn't make sense, and won't really work that well (performance))

src/layouts/tree/layout.vue
```diff
@@ -0,0 +1,144 @@
...
+    },
+    async loadItems(id) {
+      var parent = this.viewOptions.parent;
+      var items = await this.$api.getItems(this.collection, {
+        filter: { [parent]: id },
+        sort: "-" + this.viewOptions.title
+      });
+      return items.data;
+    },
+    async loadFriendItems(id) {
```
rijkvanzanten: Lets merge `parent` and `friends` and call it `relations`. You can call this `loadRelationalItems`

src/layouts/tree/layout.vue
```diff
@@ -0,0 +1,144 @@
...
+        folderDepth[id] = ids.includes(id) || (friendIds && friendIds.includes(id));
+      });
+
+      return folderDepth;
+    },
+    openItem(id, friend) {
+      var collection = friend ? this.viewOptions.friends : this.collection;
+      var projectKey = this.$store.state.currentProjectKey;
+      this.$router.push(`/${projectKey}/collections/${collection}/${id}`);
+    }
```
rijkvanzanten: Can we render the individual item (leaf) as a `<router-link>` instead? That gives us the added a11y benefit of an `<a>` tag

src/layouts/tree/options.vue
```diff
@@ -0,0 +1,219 @@
...
+  data() {
+    return {
+      sortList: null,
+      titleValid: true,
+      showAllCollections: false
+    };
+  },
+  computed: {
+    buttonLabel() {
+      return this.$t("layouts.tree.showAllCollections");
```
rijkvanzanten: `layouts.tree.showAllCollections` => `layouts.tree.show_all_collections`

src/layouts/tree/options.vue
```diff
@@ -0,0 +1,219 @@
...
+  },
+  methods: {
+    changeFriendsCollection(event) {
+      //needed to reset the "friends-field" select
+      this.viewOptions.friends = null;
+      this.viewOptions.friendsField = null;
+      this.viewOptions.friendsTitle = null;
+      setTimeout(() => {
+        this.setOption("friends", event);
+      }, 1);
```
rijkvanzanten: Why do we need the setTimeout?

src/layouts/tree/options.vue
```diff
@@ -0,0 +1,219 @@
...
+      for (var i = 0; i < keys.length; i++) {
+        if (typeof obj[keys[i]] === "object") {
+          var subKeyList = this.getKeys(obj[keys[i]]);
+          for (var k = 0; k < subKeyList.length; k++) {
+            subKeys.push(keys[i] + "." + subKeyList[k]);
+          }
+        }
+      }
+      return [...keys, ...subKeys];
+    },
```
rijkvanzanten: This is a bit hard to read / follow. Can you add a comment detailing what this does and why?

src/layouts/tree/Leaf.vue
```diff
@@ -0,0 +1,86 @@
...
+  },
+  data() {
+    return {};
+  },
+  computed: {
+    title() {
+      return this.item[this.parent.viewOptions.friendsTitle];
+    },
+    icon() {
+      return this.parent.viewOptions.friendIcon;
```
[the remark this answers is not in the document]
Nitwel: I'll call it root*

src/layouts/tree/Leaf.vue
```diff
@@ -0,0 +1,86 @@
...
+
+  .dot {
+    z-index: 1;
+    position: absolute;
+    width: 10px;
+    height: 10px;
+    border: 2px solid var(--darkest-gray);
+    background-color: var(--white);
+    border-radius: 50%;
+    float: left;
```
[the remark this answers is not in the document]
Nitwel: Oh, just leftovers from testing stuff.

src/layouts/tree/layout.vue
Nitwel: Lets make this an option in the settings instead. e.g. "prefetch items" or "automaticly load full tree"

src/layouts/tree/layout.vue
```diff
@@ -0,0 +1,144 @@
...
+import Tree from "./Tree.vue";
+
+export default {
+  components: {
+    Tree
+  },
+  mixins: [mixin],
+  data() {
+    return {
+      self: this,
```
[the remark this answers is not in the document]
Nitwel: actually wrong because you cant pass "this" inside the template section. ^^

src/layouts/tree/options.vue
Nitwel: leftover of old code

src/layouts/tree/Tree.vue
```diff
@@ -0,0 +1,202 @@
...
+        this.loading = true;
+      }
+      this.open = !this.open;
+
+      var self = this;
+      if (this.allLoaded) {
+        this.loading = false;
+        return;
+      }
+      this.parent
```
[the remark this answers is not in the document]
Nitwel: Its called `leaf.vue` because the items cant have other sub-items.

nicam:
Awesome feature. If it were possible to combine this with many 2 any relations it would enable directus to create any sort of navigation or content structure.

Nitwel:
Well nice Idea 👍 but I think we should keep it simple and just support m2o relations. The layout is quite complicated already.

rijkvanzanten:
When changing the view to Tree from any other one shows a white screen + a bunch of errors in the devtools:

image

Can we default the layout options @Nitwel? This is a little unfriendly to use

rijkvanzanten:
When adding a new relationship, I get the error

`Error: Invalid field "-undefined" in "services"`

and

`TypeError: "oldValue.connections is undefined"`

rijkvanzanten:
Also lets add a default state with a warning message like "You need to setup this layout to continue" or something along those lines. We can use a `<v-notice>` for that

rijkvanzanten:
Closing this in favor of moving it to and cleaning it up in [app-next](https://github.com/directus/app-next)

PR #2387, opened 2020-01-06
Feature: M2M default values

arboehme:
Hi,

for one customer requirement I've implemented the configuration option for M2M interface to be able to define default values for the target collection.

Currently it's only basic implementation, so admin can just define comma-separated list of the IDs from the target collection and those will automatically fetched and presented for the editor of the main collection in the create use case.  

I can imagine, that instead of comma separated list some kind of content selector can be implemented later. But for my requirement the text input is sufficient. 

Please take a look, whether the implementation is ok. 

Thanks and best regards!


rijkvanzanten:
The only thing that's a bit less-than-ideal is the fact that this stages the value on load, meaning that the app will always trigger the "are you sure you want to leave this page" warning on navigation, even if you didn't edit anything yet.

I'm wondering if we might be able to handle this on the API by relying on the regular `default_value`. I believe the API currently saves the default_value on the column in MySQL directly, which wouldn't work for a m2m, as it's not an actual column in the DB..........

This default value also wouldn't be set automatically if you're using the API, which is something the "regular" default value does. @directus/api can we consider adding "native" support for o2m/m2m default values?

rijkvanzanten:
We should handle this on the API side, just as the default value for any other field. Closing this in favor of https://github.com/directus/api/issues/1685

PR #2415, opened 2020-01-15
changed interface for multi-select to key-value (#2357)

### LINKED ISSUE #2357: Multi-select UI seems broken in 8.3.0
Directus version 8.3.0 (migrated DB from 7.x.x and I haven't had the same issue with 7.11.0)

I tried to create a multi-select field with the following options:
Screen Shot 2019-12-20 at 1 30 58 PM

But when I go on my collection, it displays like this:
Screen Shot 2019-12-20 at 1 31 24 PM

When I connect to the MySQL DB, I can see that the multi-select field can be correctly updated as I can see a `,0,1,` value when I update it from the collection UI, with the expected value.  But the display of the options is wrong, and not updated (no value selected).

dwasilew:
Hello, 

i changed the interface to key-value. This fixes #2357 
I tested it locally and also fixed one or two typos.

BR

### DIFF
--- a/src/interfaces/README.md
+++ b/src/interfaces/README.md
@@ -1,11 +1,11 @@
 # Interfaces
 
-Interfaces are the way of inputting, editing and viewing data. The most easy way to understand interface is a datepicker. It allows you to represent the date in specified format & you can validate the inputs. It renders the date in better way by providing a diffrent UI than just a normal text.
-In Database terms, interface reprents the column's value.
+Interfaces are the way of inputting, editing and viewing data. The most easy way to understand interface is a datepicker. It allows you to represent the date in specified format & you can validate the inputs. It renders the date in better way by providing a different UI than just a normal text.
+In Database terms, interface represents the column's value.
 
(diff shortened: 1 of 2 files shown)


benhaynes:
Thanks @dwasilew! We'll check this out asap...

lbrpdx:
Has this been integrated into 8.4.0? Because I can still reproduce #2357 on 8.4.0. Thank you.

rijkvanzanten:
No, this will be released on next release

lbrpdx:
Ok, thanks for the update. I was confused because you put it in "To do in v8.4.0" and then removed it... not sure if the removal was because it's now "Done" or postponed. Now I have the answer. Thanks again.

lbrpdx:
I can still reproduce #2357 on 8.5.5 for what it's worth.

benhaynes:
@lbrpdx you should comment on the ticket itself so we can re-open. Pleaseprovide any information you think would be useful in replicating the problem again.

Issue #2437, opened 2020-01-21
Bug Button Group?


andgar2010:
Version Directus: 8.4.0

Hi Team Directus,
I find that a button interface does not display correctly because I set parameters. It is not bug or misconfigured? 

![image](https://user-images.githubusercontent.com/19521054/72826784-7adefc80-3c47-11ea-859a-af167c7dbba7.png)

![image](https://user-images.githubusercontent.com/19521054/72826696-51be6c00-3c47-11ea-9f18-d3d713019e4a.png)

![image](https://user-images.githubusercontent.com/19521054/72826762-70246780-3c47-11ea-859c-34f80a26d121.png)

lluishi93:
When is this bug going to be released? @rijkvanzanten thank you

rijkvanzanten:
The bug itself was already released, the fix will be released in the next release, which is most likely around the end of this week 🙂

PR #2442, opened 2020-01-21
Fix button group rendering

rijkvanzanten:
Fixes https://github.com/directus/app/issues/2437


rijkvanzanten:
And now for the moment supreme; creating a dedicated `v-button-group` component

PR #2477, opened 2020-01-27
Work around Netlify insecure warning

rijkvanzanten:
I included a script inside the preview head for Storybook that allows us to use Vue local devtools running in Electron to tap into the Storybook iframe. Unfortunately that sets off Netlify's automated mixed http/https warnings which cause the CI checks to fail. This should work around that, while still providing the correct functionality.


rijkvanzanten:
According to Netlify support on email, this is not something they're looking to address anytime soon:

Me: 

> ... Can I disable that check somehow? Are there any work arounds to ignore that one specific script? Thanks!

Netlify: 

> Not possible at this time.  Using an https link is the only way to solve it (besides doing something like string trickery such as `"htt"+"p://localhost"` in javascript).
> 
> I've added your voice to the open feature request, but our CTO has been pretty firm in saying WONTFIX on it - wanting instead people to fix their links.

rijkvanzanten:
While a bit nasty, it seems to work: https://github.com/directus/app/pull/2477/checks?check_run_id=411256449

Issue #2504, opened 2020-01-31
Feature request: allow tiny_mce plugin options


dur41d:
TinyMCE allows setting plugin options like `paste_data_images` or `directionality: "rtl"`. Currently there is no way to set these options in the wysiwyg interface. I suggest adding a `plugin_options` json text area in the interface options tab to allow the user to specify these options in json. I already have PR for this.

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

rijkvanzanten:
This was added in https://github.com/directus/app/pull/2506

PR #2506, opened 2020-01-31
Allow tiny_mce plugin options #2504

dur41d:


rijkvanzanten:
This allows you to override any of TinyMCE's options. I'll rename it to something broader to reflect that

dur41d:
If I put the spread `...pluginOptions` in the beginning of the object then it does not override but only adds. What do you think?

rijkvanzanten:
Hmm I think it makes sense that the option would override what we provide before. It's a very "advanced user" thing to override these settings, so I would assume that you know what you're doing at that point

rijkvanzanten:
The only thing I'd like to change is the name "plugin options" to be something broader like "tinymce options". We could even add a link to the tinymce docs in the option note

PR #2535, opened 2020-02-04
Use imported lodash functions instead of relying on global

rijkvanzanten:
Makes testing easier, prevents typescript and other analysis tools from complaining


rijkvanzanten:
turns out one of these was missing a single import. must've been cached or smth on my end 😒

nachogarcia:
It's not solved yet in 8.5.3 as we referenced in slack 
image

PR #2542, opened 2020-02-05
fix avatar input image height

everyx:
image
vs 
image


andgar2010:
Have you implemented fix bug in Directus 8.5.4?

image

image

rijkvanzanten:
Looks like @everyx only pushed the fix for the `owner` interface. I'll push the same change for the user-modified one now too 🙂

PR #2571, opened 2020-02-06
Remove lodash dependency

jooola:
- Replace `clone(source)` with `Object.assign({}, source)`;


rijkvanzanten:
Lodash is currently offered to custom extensions through `window._`; removing Lodash is a breaking change

rijkvanzanten:
Also, I don't think there's a need to completely get rid of Lodash at this point either, but happy to discuss.

jooola:
The idea is to get ride of the dependency, but I didn't knew it was used by custom extensions.

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

rijkvanzanten:
Ha, I'm aware that you don't _need_ Lodash, but in the same vein we also don't _need_ Vue, or any other framework or build tool for that matter. The use of Lodash here is because we _want_ it, not _need_ it. For example, 

```js
let file = clone(item);
```

is way easier to mentally parse than

```js
let file = Object.assign({}, item);
```

Sure, there's a couple examples, like `filter` on an array, where the native implementation is so similar to native JS that it doesn't make sense to use an external function to do it, but there are also a lot of instances, like `mapValues` on an Object where a native implementation is _way_ more annoying to work with than the corresponding Lodash method.

Ironically enough, it literally says the following in the link you sent:

> Note that most native equivalents are array methods, and will not work with objects. If this functionality is needed and no object method is provided, then Lodash/Underscore is the better option.

PR #2572, opened 2020-02-06
Contextual Menu with Header Button Positioning Issue

jaruesink:
- Fixes an issue where the Contextual Menu can run off the screen when associated with the Header Button
- Updated a styling issue where expected styles were not being applied to the icon within the Contextual Menu
- added tabs rule to .pretterrc so spaces do not override on save.


rijkvanzanten:
Thanks @jaruesink!

PR #2573, opened 2020-02-06
Lazy load routes components

jooola:
This allow us to only load the strict working minimum as first files and progressively load the components afterwards.


jooola:
I saw the comment in `vue.config.js` https://github.com/directus/app/blob/master/vue.config.js#L22-L23

And I was wondering why JavaScript files should be counted in the rate limiting ?

rijkvanzanten:
@jooola all interface are already lazy loaded, by default, Vue CLI puts all those external lazy loaded files in the head with `prefetch` tags, making the browser fetch those files before they're needed. Even though you're right in saying that you should consider whitelisting static assets from rate limits, that's out of our control. Opening the app fresh used to cause upwards of 250 requests at once, which was setting off server's rate limits and anti-DDoS protections. (It also kinda negates the perf gains of lazyloading as you're blocking the thread continuously with all those requests and resolves)

rijkvanzanten:
Lets keep the following routes imported "non-async":

```
import Collections from './routes/collections.vue';
import Items from './routes/items.vue';
import FileLibrary from './routes/file-library.vue';
import Item from './routes/item.vue';
import Login from './routes/login.vue';
```

We know for a fact that people will see the login page, and there is a very high probability they will load collections, items, and the file-library in regular use after that. More specific things like settings etc make sense to be loaded async

jooola:
> Even though you're right in saying that you should consider whitelisting static assets from rate limits, that's out of our control. Opening the app fresh used to cause upwards of 250 requests at once, which was setting off server's rate limits and anti-DDoS protections. (It also kinda negates the perf gains of lazyloading as you're blocking the thread continuously with all those requests and resolves)

See directus/api#1741 My approach might be a bit dumb, but the static files aren't rate limited since the webserver directly serves them without running any PHP. It might have been the case in earlier days, or I am missing something, but we are safe to reenable the prefetch plugin.

rijkvanzanten:
> or I am missing something

I think you forgot to consider rate limits on the webserver / load balancer level 🙂

PR #2577, opened 2020-02-07
Reenable webpack prefecth plugin

jooola:
This seemed like the next thing to do after #2573
Fixes https://github.com/directus/api/issues/1741

I didn't get any answer on the static files rate limiting so I might be missing something here.


rijkvanzanten:
Like I said in [my answer](https://github.com/directus/app/pull/2573#issuecomment-583193950) _webservers_ can have rate limits on the load balancer or proxy level, negating whether or not PHP is targeted. Enabling prefetch for every single lazy loaded component will easily cause around 250 requests to be made immediately, which is enough to trip most webserver level rate limiters.

Issue #2589, opened 2020-02-09
Translations content doesn't show when viewed via relation


pgegenfurtner:
Setup:

- "Brand" Collection with a translation interface ("brand_transltions" collection)
- "Project" Collection with a many-to-many relation to the "brand" collection.

In "Brand" I can create my items with translations and they show up correctly.
Now, under "Project" I add my created brands. When I view them (in the modal) the translations don't show up.

Directus Version: 8.5.4, fresh install and translations interface setup according to the docs.

Here is a sql dump: https://www.dropbox.com/s/kmbjyynvgr72o76/translations-problem.sql?dl=0

zerbfra:
Same bug here

zerbfra:
Hi @benhaynes  the PR here from my repo should fix the behavior.
Some notes about it:

- I get the system language and I split the first part (I don't know if it is correct) so we can get the translation "slug"...for instance `en-US` becomes `en`
- If no translation found for the current system language fallback to the first one found
- The getItem api call should get all the data to get the translation fields (and not only the id)

A note on usage: in the relation field we need to specify the field to access. I modeled it with {{translation_field.field_to_display}} and the language is computed.

Let me know what you think

zerbfra:
I made some refactor on it ;) using it right now and seems ok

pgegenfurtner:
@zerbfra Great that you took some time to look into that! Thanks!

I tested your PR but it doesn't seem to do anything on my end.
Do I need to change my configuration?

zerbfra:
> @zerbfra Great that you took some time to look into that! Thanks!
> 
> I tested your PR but it doesn't seem to do anything on my end.
> Do I need to change my configuration?

Yes, you should change in the "Project" entity, in the related field to "Brand" the  dropdown template as follows: `{{name_of_the_translation_field.field_to_display}}`

pgegenfurtner:
It doesn't work for me. 

<img width="701" alt="Bildschirmfoto 2020-02-11 um 09 55 17" src="https://user-images.githubusercontent.com/19979167/74222399-b3a54b00-4cb4-11ea-9d33-20b999265ff4.png">

The translated fields don't show up, even though they are there and visible when I go directly in the brand collection

zerbfra:
Did you get my latest version? Can you send here how did you setup the relation?

pgegenfurtner:
Yes, commit 8475bfa5 ('refactor').

[...]

PR #2592, opened 2020-02-10
Fixes the dropdown options for translated collections

zerbfra:
This should fix the issue #2589

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,21900 @@
+{
+  "name": "directus",
+  "version": "8.5.4",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
+      "requires": {
+        "@babel/highlight": "^7.8.3"
+      }
+    },
+    "@babel/compat-data": {
+      "version": "7.8.5",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.8.5.tgz",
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
I don't think the many to one interface should be responsible for handling the languages 🤔 Also, the possible languages of the translations interface are user configurable, so relying on the system language won't always work

zerbfra:
Ok, let me explain where the issue is.
At this point, in the relation fields I will have the dropdown showing empty strings.
Let me explain with an example.

I have the `product_line` table with no fields but only a translation fields named `data` (related to a `t_product_line` table).
In the `t_product_line` table I have the `name` field that is translated in two languages (`it` and `en`).

So, now I have a `category` table with a `category_product_line` field that points in a many to one relation to the `product_line` table.
Now, in the dropdown field `category_product_line` I'll get only empty strings. So I don't know what to select!

With this PR I get the translation in the dropdown. I get the one is specified in the Directus admin panel  ("Default Locale") or if there's no such translation in the `t_product_line` it will fallback to the first one available.

zerbfra:
Additionally, I saw that in the list of items `product_line` I get the translation value for the `Default Locale`..so why not displaying the value in the dropdown?

zerbfra:
@rijkvanzanten what do you think?

zerbfra:
Any feedbacks? @rijkvanzanten

rijkvanzanten:
Yeah the same as I mentioned before still applies.. The many to one shouldn't have to care about translations, as it's not related to translations in any way itself.

As for the fix, I don't think this actually fixes #2589, as that seems to be a problem with viewing translations within the edit context of a o2m modal, not the dropdown of the M2O. This change will show translated values in the m2o dropdown by trying to use the system language

zerbfra:
Yes, it is correct. The PR is unrelated but the problem still exists.
Can I help someway to fix? I have a big project undergoing and I'm using Directus as a CMS, I made my fork to handle this issue but I can help you to fix everything correclty for an official patch for Directus

rijkvanzanten:
@zerbfra I think we could / can / should refactor the M2O interface to render display components inline instead of just text, so we can have the translations display component handle the rendering of the correct values.

As for the problem in #2589, we'll have to investigate what's going on in the nesting of the translations interface inside of a o2m to know how to fix it. If you want, you can hit me up [on Slack](https://directus.chat) to discuss a little easier 🙂

rijkvanzanten:
Closing this in favor of the solution added in [v9](https://github.com/directus/directus/releases/tag/v9-alpha.1). In v9, the m2o will render the display template using actual displays, which in turn means that the display for the translations interface can be in charge of what to render nested. I've also thrown in support for deeply nested relations in this display template, so you can use the value that's stored in another relational field in the translations

PR #2633, opened 2020-02-27
Ability to make repeater rows duplicable

nicolas-t:
I have a usecase where it would be very useful to be able to duplicate rows

Repeater field creation/edition (`false` by default) : 
image

Item creation/edition : 
image

### DIFF
--- a/src/interfaces/repeater/input.vue
+++ b/src/interfaces/repeater/input.vue
@@ -15,11 +15,13 @@
 				:fields="repeaterFields"
 				:inline="inline"
 				:template="options.template"
+				:duplicable="options.duplicable"
 				:open="open === index"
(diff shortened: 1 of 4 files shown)


benhaynes:
Thanks, this looks awesome! My main feedback would be to use a context menu if there's more than one icon to the right. That's the three dot icon that opens up a menu when clicked. @rijkvanzanten ?

nicolas-t:
Thanks @benhaynes, I just pushed a new version with the contextual menu, it's better like this.
image

benhaynes:
Awesome work!! And this only shows for 2+ icons?

nicolas-t:
Yes, nothing change if you dont opt-in for "duplicable rows" in the collection settings.

PR #2639, opened 2020-03-03
Fix code interface syntax highlighting

nicolas-t:
Hello,

If I add a code interface (javascript mode) to my collection, when editing/creating an item I get no syntax highlighting.
This is because `mode` is undefined here : 
https://github.com/directus/app/blob/master/src/interfaces/code/input.vue#L89

`mode` is a computed so it's undefined before `data()` is executed.

With this pattern you also have to use some imperative code in a watcher to keep the data object in sync.

Instead I used a computed.
I hope this doesn't introduce any side effects.

### DIFF
--- a/src/interfaces/code/input.vue
+++ b/src/interfaces/code/input.vue
@@ -76,17 +76,14 @@ export default {
 		return {
 			lineCount: 0,
 
-			cmOptions: {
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Good catch!

PR #2644, opened 2020-03-04
set newItem property to true on row when a new row is added

ttcremers:
I'm not entirely sure that this fits the pattern of how you'd normally manage these types of values. I had considered that this might needs to be set in the initial field definition of the slug field but that didn't feel right. 

Thanks!


rijkvanzanten:
LGTM!

PR #2650, opened 2020-03-06
Allow no matches in getFieldsFromTemplate

nicolas-t:
Hello,

If I create a repeater row with a template like this : 
image

I get this error 
```
TypeError: Cannot read property 'map' of null
    at getFieldsFromTemplate (get-fields-from-template.js?1581:18)
```

Indeed `getFieldsFromTemplate` only works when the `string` argument contains mustaches.

My fix : 
If no mustache-matches are found in `string` we return an empty array.


rijkvanzanten:
Thanks @nicolas-t! 

Technically speaking, it would be quite useless to set up the display template without using any field, but a good thing to account for anyway 🙂

PR #2660, opened 2020-03-16
fix: repeater row should propagate width attribute to v-ext-input

ttcremers:
Shouldn't the width attribute from the field definition always be propagated as it can be set in the repeater field model when creating/updating? Seems like `half` is a very sane default, can't believe I never noticed this before.

image


rijkvanzanten:
Yup! Good catch. The actual form grid is controlled by the parent form, so it's easy to miss. This width prop is only used by a couple interfaces to slightly alternate how they render based on the available space they're given.

Issue #765, opened 2018-09-08
Export/Import Schema and Settings


benhaynes:
#### :rotating_light: MIGRATED FROM REQUESTS.GETDIRECTUS.COM :rotating_light: 

# Ability to export tables, columns and interfaces setup
### :+1: = 2
*Created 10 months ago by @rijkvanzanten*
 

-----
### @rijkvanzanten – :+1: (10 months ago)
Not exporting raw data, but the actual Directus setup

rijkvanzanten:
Ref https://github.com/directus/api/issues/31

benhaynes:
Good one...

rijkvanzanten:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/app/projects/45) project board.

**This issue being closed does not mean it's not being considered.**

yagobski:
Hi Any progress on this feature. How can i have export button to export the results of filtering in CSV format?

benhaynes:
Unless there's some external help, our staff is full-steam ahead on the API and App refactors for the next few months. We'll likely include those features in the new API port, but again, that is probably 8-12 weeks away. We're looking to add new devs to the team to speed that up though. #oss

PR #2747, opened 2020-04-20
Export/Import Schema and Data #765 (not Settings)

nathannau:
Allow export and import of schema and data (but not yet settings)


nathannau:
Bugged commit. Sorry

PR #2748, opened 2020-04-20
Allow external scripts access to modules used within Vue...

philleepflorence:
Docs to be updated and blog post written upon PR approval.


rijkvanzanten:
Interesting solution! I might enhance this with extra info like version etc

PR #2749, opened 2020-04-21
Export/Import Schema and Data #765 (not Settings)

nathannau:
Allow export and import of schema and data (but not yet settings)


rijkvanzanten:
This is a really really cool solution!

That being said I do think we should have the actual export / import logic on the API level instead of in the app, to ensure you can use it without having to rely on the app to do it. Is that something you'd be willing to take on @nathannau?

PR #2758, opened 2020-04-23
fix translation on popup and inscrease debounce

cavaon-wayne:
The idea is that in most of the case the translation data have been fetched and it can be access from the `values` passed from its parent component or interface, so there is not need to use `newItem` for checking the values. 

Regarding to the issue on modal, I think the issue is that it didn't emit the whole value it was stored.

changing the debounce as I found that in the item-select model, if I typed "syd" and tried to delete "yd" and typed other "xx", the result will looks very funny. Thinking most of the case, around 1 seconds waiting still fine for the user.

### DIFF
--- a/src/components/item-select/item-select.vue
+++ b/src/components/item-select/item-select.vue
@@ -254,7 +254,7 @@ export default {
 
 		this.fetchItems();
 
-		this.setSearchQuery = debounce(this.setSearchQuery, 550);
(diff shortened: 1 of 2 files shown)


cavaon-wayne:
Found a bug when trying to set a new translation into the item, so fixed it on this commit: https://github.com/directus/app/pull/2758/commits/1f63cc8e21022ff044c704d820ca9857d5f70f56

PR #2773, opened 2020-04-29
Feature: Add new filter option for m2m interface 

arboehme:
Add new filter option for m2m interface to be able to configure the default filter for the relations in the collections.

This allows to specify comma separated list of filters as described in https://docs.directus.io/api/query/filter.html, e.g. 

`filter[category][eq]=development,filter[category][eq]=consulting`

The item list provided for selection is filtered then with help of these filters.

Following improvements I can imagine:
- Instead of using the text input, the repeater can be used to specify the filters (similar to translations)
- Since I'm not familiar with i18n concept of options, I put only english and german versions in these pull request. Other languages can be extended. 

Best regards

### DIFF
--- a/src/components/item-select/item-select.vue
+++ b/src/components/item-select/item-select.vue
@@ -284,15 +284,12 @@ export default {
 			const params = {
 				limit: 200,
 				offset: options.offset,
-				meta: '*'
+				meta: '*',
+				...(this.filters.length > 0 ? formatFilters(this.filters) : {})
 			};
 
 			if (this.searchQuery.length > 0) {
 				params.q = this.searchQuery;
-			}
-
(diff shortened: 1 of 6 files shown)


rijkvanzanten:
This is awesome @arboehme! Thanks 🙂 

Do you think we can update the filter option from a text-input to a repeater? That way it's a little more user friendly to setup 💪

arboehme:
@rijkvanzanten  For sure, the repeater would be much more user friendly. Unfortunately, I simply don't have time to enhance this... :(

Do you have an example, where the repeater is used as an interface option? If I have some time,  maybe I can use it as reference to replace the text input field with the repeater

PR #2777, opened 2020-04-30
Feature: Add toggle all check box in item-select component

arboehme:
To select all items with one click:

Bildschirmfoto 2020-04-30 um 19 03 29

If only some items are selected, the checkbox is shown as follows:

Bildschirmfoto 2020-04-30 um 18 57 54

PS: The "load more" button is considered as well, so the check box change the state, if more items are loaded


src/components/item-select/item-select.vue
```diff
@@ -48,7 +48,23 @@
...
-					<span />
+					<span>
+						<v-icon
+							@click="toggleAll()"
+							:name="
+								value.length == 0
+									? 'check_box_outline_blank'
+									: value.length == items.length
+									? 'check_box'
+									: 'indeterminate_check_box'
```
rijkvanzanten, line 59: Not too big of a fan of the double-ternary, but it beats having another computer prop I suppose

arboehme: Yes,  that's the reason why I used this operator :) Since there are only 3 options, I didn't want to add the new property

PR #2786, opened 2020-05-07
Fix infinite scroll

cavaon-wayne:
Issue 1: Current Infinite Scroll only load 2 page
Issue 2: When use apply a filter and count number showed on the search box still the same as the one without the filter

So I create two commits to fix them.


benhaynes:
Thanks @cavaon-wayne!! We're going to be switching to proper pagination in v9, but this is a great PR until then! We'll review and merge as soon as we have time

PR #2788, opened 2020-05-08
Add new Status-field as dropdown

zghib:
Hey guys here I've decide to add a new status field as a dropdown. 
I still don't get why it doesn't work when I add it as a custom extension.
I think that it will be helpful to have two representations of the status field, as in my case with 6 statuses checkboxes are ugly.

### DIFF
--- a/src/interfaces/status_dropdown/display.vue
+++ b/src/interfaces/status_dropdown/display.vue
@@ -1,0 +1,20 @@
+<template>
+	<div class="no-wrap">
+		{{ currentStatus.name }}
+	</div>
+</template>
+
+<script>
+import mixin from '@directus/extension-toolkit/mixins/interface';
+
+export default {
+	mixins: [mixin],
+	computed: {
+		currentStatus() {
+			return this.options.status_mapping[this.value];
+		}
+	}
+};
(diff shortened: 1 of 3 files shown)


benhaynes:
Thanks @zghib!! We'll take a look and let you know what we find...

rijkvanzanten:
Thanks @zghib! Could you add support for some of the other status mapping settings in options as well? 

https://github.com/directus/app/blob/b73ef765f9786b755e6dc9f84aa17f3c02c34fea/src/interfaces/status/meta.json#L28-L74

You can ignore the colors, but it's pretty important that you're able to configure soft-delete and published, as that actually changes the API output

rijkvanzanten:
And fun fact, we were actually just about to add this as an interface in the upcoming v9 too 🙂 

CleanShot 2020-05-18 at 12 12 55

zghib:
> And fun fact, we were actually just about to add this as an interface in the upcoming v9 too 🙂
Great, we are not in a rush with this, so waiting for v9, any ETA for v9?

rijkvanzanten:
We actually released the first alpha of it last friday! 😄

PR #2824, opened 2020-05-25
Added a badge-style interface for the ManyToOne field.

tuxnese:
I have replicated the badge style of the status field in the ManyToOne interface. 
It is possible to define the background and text color of each referenced value by using to dedicated color fields in the referenced table. 

Here is an example.
Screenshot from 2020-05-25 15-53-10

An alternative style could be to have a transparent background, black text and colored border.

I think that is PR would improve the user experience. Feel free to modify the code and the style.


benhaynes:
Nice!! I've been wanting this for a while, thank you!

In the upcoming v9, we're handling all this a bit differently: The "Interface" (presentation on the Item Detail page, mostly for interaction) is now decoupled from the "Display" (presentation elsewhere, read-only). For example, this means we can have several interfaces that all use the "Tags" display on the Item Browse page... or multiple interfaces that use the "Badge" display.

The one issue @rijkvanzanten is working on now is how to handle instances when the display needs info from the interface options... such as with Status Mapping Options (where the colors/names are).

Happy to chat about this more on our Slack if you'd like! https://directus.chat

tuxnese:
The separation between Interfaces and Displays looks like a very good idea.

PR #2876, opened 2020-06-16
safari no longer crashes when parsing file created_on date field. 

jroxendal:
fixed error which caused safari to crash when parsing image.uploaded_on date string. More info here:
https://stackoverflow.com/questions/4310953/invalid-date-in-safari

### DIFF
--- a/src/interfaces/file/input.vue
+++ b/src/interfaces/file/input.vue
@@ -136,7 +136,7 @@ export default {
 					.pop()
 					.toUpperCase() +
 				' • ' +
-				this.$d(new Date(this.image.uploaded_on), 'short')
+				this.$d(new Date(this.image.uploaded_on.replace(/-/g, '/')), 'short')
 			);
 		},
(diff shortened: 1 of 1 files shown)


jroxendal:
The error i'm fixing shows up here with Safari:
https://demo.directus.io/admin/#/thumper/users/1

and causes the avatar image loading to crash. Safari throws `RangeError: Date value is not finite` from e.g. `new Date("2019-12-28 20:33:25")`

jroxendal:
Arguably the API shouldn't be sending datestrings that cannot be parsed by every browser, but I'll leave it up to you whether to fix it here or in the API!

rijkvanzanten:
@jroxendal You're absolutely right. We're actually standardizing all dates everywhere to use ISO8601 to prevent these kind of mismatches moving forward. That's a bit of a big undertaking though, so this fix is very welcome in the meantime

PR #2895, opened 2020-06-30
Allow BIGINT db type for m2o fields

apiraino:
Hello!

In our usecase we have DB fields used as foreign key exceeding the size of MySQL INT type and we therefore use [BIGINT](https://dev.mysql.com/doc/refman/5.7/en/integer-types.html).

This patch allows Directus app to use BIGINT type fields when creating a many-to-one relationship between two tables. The default type INT stays as before, so this change should be fully retrocompatible.

thanks @rijkvanzanten for a review :)

cc: @herrfinke


apiraino:
thanks :+1:

Issue #2410, opened 2020-01-14
WYSIWYG interface does not show link modal when used within oneToMany


Wendenburg:
# Bug Report

## Steps to Reproduce

1. create a collection 'Tabs' with an oneToMany interface
2. create a collection 'TabsContent' containing something like a title and a **WYSIWYG interface** The 'TabsContent' has to be related to 'Tabs' by the oneToMany interface
3. create an entry in 'Tabs', press the 'add new' button of the oneToMany interface
4. now the 'TabsContent' is shown in a modal. you can add content as normal
5. mark some content to add a link, press the 'link' icon of the WYSIWYG interface to open the link modal

## Expected Behavior

the link attributes should be shown in in modal to be edited

## Actual Behavior

since oneToMany opens the related collection in a modal itself we have two overlaying modals now. The main modal containing 'TabsContent' covers the link attributes. Also we have two grey overlays above each other

## Other Context & Screenshots

![Bildschirmfoto 2020-01-14 um 11 44 30](https://user-images.githubusercontent.com/25299148/72339011-e49a5c00-36c5-11ea-802a-90d778d7ad41.png)

reducing the z-index of **div data-v-7a42557c="" data-v-6007a4c2="" class="v-modal" style=""** to something like 1000 could be a quick fix...

![Bildschirmfoto 2020-01-14 um 12 04 31](https://user-images.githubusercontent.com/25299148/72339382-6a1e0c00-36c6-11ea-863c-be0378c739ef.png)

## Technical Details

Device: desktop
OS: Mac OS 10.13.6
Web Server: Apache, Mamp
PHP Version: 7.3.8
Database: 5.7.26
Install Method: both, cloned master branch and docker container

imrick:
The same can be observed while trying to upload an image from the editor.

**Version** 8.7.2
**Browser** Chrome
**Steps to reproduce :**
1. create a collection "page" with an oneToMany interface linked to an other collection (ex: "bloc"). Bloc should have a wysiwyg field
2. create a page entry then add a new bloc
3. In the wysiwg clic on the insert image button and the in the modal try to add image

![Capture d’écran de 2020-06-29 16-34-32](https://user-images.githubusercontent.com/1628410/86019113-7dc7ef80-ba26-11ea-8457-95c5f370a38c.png)

Changing the z-index will not resolve the issue in that case since we have 3 modals here that are not in the branch of the DOM tree :
- v-modal oneMany
  + v-modal upload file
- popin tinymce insert image

[...]

PR #2896, opened 2020-06-30
Fix WYSIWYG modals display within oneToMany (#2410)

imrick:
Hello,

The solution is not ideal but since there is no solution in TinyMCE API to change Dialog dom root element i don't see any other solution. The TinyMCE dialog is hidden during file selection preventing side effects when using it in a oneToMany modal scenario.
Tested also when WYSIWYG editor is outside a modal.


rijkvanzanten:
Thanks @imrick! TinyMCEs default markup and styles can be quite tricky to work around indeed..

Issue #2870, opened 2020-06-12
Error in wysiwyg input when used within modal


IgordeOliveira:
Hello, I'm having a problem here, when I try to edit a field of type many to many ( another collection ), and this collection contains a field of type wysiwyg ( editor ) and saved, the console points out that the editor is undefined (screenshot below), and when I do it several times, it does not save the new content written

• Version of Directus: 8.7.2
• Server, OS & Database Details: Docker, MySql
• Error Messages
![Anotação 2020-06-12 165603](https://user-images.githubusercontent.com/19624377/84542533-08db8280-acd0-11ea-96a6-469bf475a28b.png)

• Steps to Reproduce
Try editing the value of a wysiwyg editor from another collection
• Relevant Info, Screenshots, Schema & Logs
![Anotação 2020-06-12 170212](https://user-images.githubusercontent.com/19624377/84542624-3aece480-acd0-11ea-904a-d15af239fa32.png)
![Anotação 2020-06-12 170251](https://user-images.githubusercontent.com/19624377/84542625-3c1e1180-acd0-11ea-8417-9aaf288dd734.png)

🐰

imrick:
Hello, i have the same error within a oneToMany field.

**Steps to reproduce**

1. Create a collection that have a oneToMany field to an other collection
2. Add a wysiwyg field in the many collection
3. Add an item in the many collection field
4. Select an image then click "save" but without typing anything in the editor

**Analysis**

This is related to the debounce on "updateValue". The blur event is indeed trigger properly but the callback happens after the editor was destroyed.

**Possible solution**

Only the "@OnkeyUp" event should be debounced to preserve good performances. By removing debounce on the other events the problem is solved.

rijkvanzanten:
Fixed in https://github.com/directus/app/pull/2898

PR #2898, opened 2020-06-30
Fix WYSIWYG undefined editor on value update within oneToMany (#2870)

imrick:
Hello,

A solution to the issue #2870. I renamed the debouned updateValue as  "planToUpdateValue" and keep it only on the keyUp event to preserve good performance. I think that the other events should trigger value update immediatly. 
Hope you are ok with that.

Also i removed the "console.log" that i have forgotten previously. Sorry for the inconvenient.

### DIFF
--- a/src/interfaces/wysiwyg/input.vue
+++ b/src/interfaces/wysiwyg/input.vue
@@ -4,7 +4,7 @@
 			ref="editorElement"
 			:init="initOptions"
 			:value="value"
-			@onKeyUp="updateValue"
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
I don't know what happened here, but it looks like you're adding a `.patch` file with the changes you intended to make.. ?

CleanShot 2020-07-01 at 11 05 17

imrick:
My bad, just pushed the patch instead of applying it -_-
It's fixed

rijkvanzanten:
Thanks!

PR #2919, opened 2020-07-19
fix(#2911): Unable to handle newly uploaded file

### LINKED ISSUE #2911: File/files interfaces not working as expected
# Bug report
Cannot upload new files using file/files interfaces

## Steps to reproduce
1. add file/files interface to collection
2. set file as required  **(optional)**
3. create new collection's item
4. try to upload file using those interfaces
5. try to save the item
6. try to manually choose file(s) in both cases
7. and save again **(optional)**

## Expected behavior
Upload/choose file using file/files interface and save the item

## Actual behavior
1. When trying to upload new file (both **"embed"** and **"select form device"**), with file/files interfaces, those cannot be chosen as part of collection's item, but existing (already uploaded) files can. The desired file is actually uploaded in the background, but it is not selected (and metadata isn't shown). After problematic upload the same item can be selected from file library and the problem with metadata is no more.

2. When trying to save collection's item got those **error messages**: 
 `fajl: This value should not be blank.`   _(in this case file item is required)_
 `filename_disk: This value should not be blank. uploaded on: This value should not be blank. filename_download: This value should not be blank.`
 and those are fired in console, too. Item can be saved after selecting files from library, but for files interface error is fired in console again.

- **Remark:** _tried both windows and linux machines_

## Screenshots
Selecting the file:
- **file interface** _(without errors in console)_
image

- **files interface**
_upload:_
 image 
_select:_ 
image 
_result:_ **(the same file!)**
image 

## Other notices
- Maybe this issue has something to do with #2308
- The similar metadata issue found there [#1869](https://github.com/directus/api/issues/1869)
- This wasn't issue in previous Directus version 8.7.x but that version has had problems with custom user roles, so that was the main reason to change to the newest.

## Technical details
- **Directus version 8.8.0**
- **Comercial SSD linux servers** (different providers)
- **MySQL db**
- **PHP 7.3**
- **Installed with manually uploaded zip file.**

lemankk:
"id" prop doesn't exist in response payload "data.data"

Fixes #2911

### DIFF
--- a/src/interfaces/file/display.vue
+++ b/src/interfaces/file/display.vue
@@ -27,7 +27,7 @@ export default {
 			return this.filetype && this.filetype.startsWith('image');
 		},
 		icon() {
-			return getIcon(this.filetype);
+			return getIcon(this.filetype || '');
 		},
 		imageUrl() {
 			if (!this.isImage) return null;
--- a/src/interfaces/file/input.vue
+++ b/src/interfaces/file/input.vue
@@ -251,7 +251,7 @@ export default {
 			window.open(this.image.data.full_url);
 		},
 		saveUpload(response) {
-			this.image = response.data.data;
+			this.image = response.data;
 			// We know that the primary key of directus_files is called `id`
 			this.$emit('input', this.image.id);
 		},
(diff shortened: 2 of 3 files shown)


rijkvanzanten:
CI failed due to an unrelated issue in Coveralls

svassr:
Hi. When is this fix supposed to land in https://github.com/directus/directus ?

PR #2972, opened 2020-09-03
improve icon searching

cynova:
Icon searching is currently quite bad.
Using tags it can be improved immensely. This pull request also disables searching for single letter queries for performance reasons.

### DIFF
--- a/src/interfaces/icon/input.vue
+++ b/src/interfaces/icon/input.vue
@@ -8,7 +8,7 @@
 			icon-right-color="input-background-color-active"
 			icon-left="search"
 		></v-input>
-		<div v-show="searchText.length === 0" class="icons-view">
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
@cynova can you make the tags / icons JSON arrays be loaded asynchronously? I'd rather not include 270KB in the core bundle just to enhance the icon interface a little 👀

cynova:
> @cynova can you make the tags / icons JSON arrays be loaded asynchronously? I'd rather not include 270KB in the core bundle just to enhance the icon interface a little 👀

Of course, done. I'm not 100% sure whether there might be a better way to do the import as i had to add a object wrapper in tags.json to prevent webpack from converting the array to an object...

